PR #4, opened 2022-12-08
[WIP] Adds reverse lookup support (primary name)

0xChucky:
Adds on-chain reverse lookup (aka primary name)

- Users can set their primary name to a name that they own. This also forces the name to target their own address too.
- If you point your primary name to an address that isn't your own, it is no longer your primary name.
- If you mint a name while you don't have a primary name, your minted name is auto-set to be your primary name. (including minting subdomains)

Notably, if you send your primary name to someone else, it's still your primary name. This will be addressed in a follow-up PR which will allow users to be able to clear the target address of any name that points to them. Clearing the target address of a name forces it to no longer be the primary name of that target address.

### DIFF
--- a/core/Move.toml
+++ b/core/Move.toml
@@ -9,10 +9,10 @@ aptos_names_funds = "0x78ee3915e67ef5d19fa91d1e05e60ae08751efd12ce58e23fc1109de8
 
 [dependencies.AptosFramework]
 git = 'https://github.com/aptos-labs/aptos-core.git'
-rev = 'main'
+rev = 'mainnet'
 subdir = 'aptos-move/framework/aptos-framework'
 
 [dependencies.AptosToken]
 git = 'https://github.com/aptos-labs/aptos-core.git'
-rev = 'main'
+rev = 'mainnet'
 subdir = 'aptos-move/framework/aptos-token'
--- a/core/Move.toml
+++ b/core/Move.toml
@@ -9,10 +9,10 @@ aptos_names_funds = "0x78ee3915e67ef5d19fa91d1e05e60ae08751efd12ce58e23fc1109de8
 
 [dependencies.AptosFramework]
 git = 'https://github.com/aptos-labs/aptos-core.git'
-rev = 'mainnet'
+rev = 'framework-mainnet'
 subdir = 'aptos-move/framework/aptos-framework'
 
 [dependencies.AptosToken]
 git = 'https://github.com/aptos-labs/aptos-core.git'
-rev = 'mainnet'
+rev = 'framework-mainnet'
 subdir = 'aptos-move/framework/aptos-token'
--- a/core/sources/domains.move
+++ b/core/sources/domains.move
@@ -608,6 +608,11 @@ module aptos_names::domains {
         event::counter(&borrow_global<RegisterNameEventsV1>(@aptos_names).register_name_events)
     }
 
+    #[test_only]
+    public fun get_set_reverse_lookup_event_v1_count(): u64 acquires SetReverseLookupEventsV1 {
+        event::counter(&borrow_global<SetReverseLookupEventsV1>(@aptos_names).set_reverse_lookup_events)
+    }
+
     #[test(aptos = @0x1)]
     fun test_time_is_expired(aptos: &signer) {
         timestamp::set_time_has_started_for_testing(aptos);
--- a/core/sources/domains.move
+++ b/core/sources/domains.move
@@ -417,6 +417,9 @@ module aptos_names::domains {
     }
 
     public fun set_name_address(sign: &signer, subdomain_name: Option<String>, domain_name: String, new_address: address) acquires NameRegistryV1, ReverseLookupRegistryV1, SetNameAddressEventsV1, SetReverseLookupEventsV1 {
+        // If the domain name is a primary name, clear it.
+        clear_reverse_lookup_for_name(subdomain_name, domain_name);
+
         let signer_addr = signer::address_of(sign);
         let (is_owner, token_id) = is_owner_of_name(signer_addr, subdomain_name, domain_name);
         assert!(is_owner, error::permission_denied(ENOT_OWNER_OF_NAME));
@@ -527,10 +530,8 @@ module aptos_names::domains {
     public fun set_reverse_lookup(account: &signer, key: &NameRecordKeyV1) acquires NameRegistryV1, ReverseLookupRegistryV1, SetNameAddressEventsV1, SetReverseLookupEventsV1 {
         let account_addr = signer::address_of(account);
         let (maybe_subdomain_name, domain_name) = get_name_record_key_v1_props(key);
-
-        clear_reverse_lookup_for_name(maybe_subdomain_name, domain_name);
-        set_reverse_lookup_internal(account, key);
         set_name_address(account, maybe_subdomain_name, domain_name, account_addr);
+        set_reverse_lookup_internal(account, key);
     }
 
     /// Clears the user's reverse lookup.
--- a/core/sources/domains.move
+++ b/core/sources/domains.move
@@ -293,16 +293,18 @@ module aptos_names::domains {
     /// Forcefully set the name of a domain.
     /// This is a privileged operation, used via governance, to forcefully set a domain address
     /// This can be used, for example, to forcefully set the domain for a system address domain
-    public entry fun force_set_domain_address(sign: &signer, domain_name: String, new_owner: address) acquires NameRegistryV1, SetNameAddressEventsV1 {
+    public entry fun force_set_domain_address(sign: &signer, domain_name: String, new_owner: address) acquires NameRegistryV1, ReverseLookupRegistryV1, SetNameAddressEventsV1, SetReverseLookupEventsV1 {
         force_set_name_address(sign, option::none(), domain_name, new_owner);
     }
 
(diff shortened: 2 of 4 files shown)


core/sources/domains.move
```diff
@@ -135,7 +149,19 @@ module aptos_names::domains {
         token_helper::initialize(account);
     }
 
-    fun register_domain_generic(sign: &signer, domain_name: String, num_years: u8) acquires NameRegistryV1, RegisterNameEventsV1, SetNameAddressEventsV1 {
+    public fun init_reverse_lookup_registry_v1(account: &signer) {
+        assert!(signer::address_of(account) == @aptos_names, ENOT_AUTHORIZED);
```
movekevin, line 153: nit: Wrap ENOT_AUTHORIZED to be consistent with other error codes.

core/sources/domains.move
```diff
@@ -232,6 +256,16 @@ module aptos_names::domains {
...
         table::upsert(&mut aptos_names.registry, name_record_key, name_record);
 
+        let account_addr = signer::address_of(sign);
+        let reverse_lookup_result = get_reverse_lookup(account_addr);
+        if (option::is_none(&reverse_lookup_result)) {
+            // If the user has no reverse lookup set, set the user's reverse lookup.
+            set_reverse_lookup(sign, &NameRecordKeyV1 { subdomain_name, domain_name });
+        } else if (option::is_none(&subdomain_name)) {
+            // Automatically set the name to point to the sender's address
+            set_name_address_internal(subdomain_name, domain_name, signer::address_of(sign));
```
movekevin, line 266: Why do we only set the target address on the domain if the user already has a primary name? Shouldn't this always be set?

core/sources/domains.move
```diff
@@ -438,6 +493,39 @@ module aptos_names::domains {
...
+        } else {
+            option::none()
+        }
+    }
+
+    fun set_reverse_lookup_internal(account: &signer, key: &NameRecordKeyV1) acquires ReverseLookupRegistryV1, SetReverseLookupEventsV1 {
+        let account_addr = signer::address_of(account);
+        let (maybe_subdomain_name, domain_name) = get_name_record_key_v1_props(key);
+        let (is_owner, _) = is_owner_of_name(account_addr, maybe_subdomain_name, domain_name);
+        assert!(is_owner, ENOT_AUTHORIZED);
```
movekevin, line 517: nit: Wrap error code.

core/sources/domains.move
```diff
@@ -438,6 +493,39 @@ module aptos_names::domains {
         };
     }
 
+    public fun set_reverse_lookup(account: &signer, key: &NameRecordKeyV1) acquires NameRegistryV1, ReverseLookupRegistryV1, SetNameAddressEventsV1, SetReverseLookupEventsV1 {
```
movekevin, line 496: Can you add documentation for this public function?

core/sources/domains.move
0xChucky: Done

core/sources/domains.move
0xChucky: Done

core/sources/domains.move
0xChucky: Done

core/sources/domains.move
0xChucky: this part is confusing - set_reverse_lookup() also sets the target address of the domain to be the user's address

core/Move.toml
```diff
@@ -9,10 +9,10 @@ aptos_names_funds = "0x78ee3915e67ef5d19fa91d1e05e60ae08751efd12ce58e23fc1109de8
 
 [dependencies.AptosFramework]
 git = 'https://github.com/aptos-labs/aptos-core.git'
-rev = 'main'
+rev = 'mainnet'
```
movekevin, line 12: Why change this from main to mainnet?

core/sources/domains.move
```diff
@@ -71,6 +71,14 @@ module aptos_names::domains {
         registry: Table<NameRecordKeyV1, NameRecordV1>,
     }
 
+    struct ReverseLookupRegistryV1 has key, store {
```
movekevin, line 74: Can you add more documentations for these structs?

core/sources/domains.move
```diff
@@ -71,6 +71,14 @@ module aptos_names::domains {
...
+        registry: Table<address, NameRecordKeyV1>
+    }
+
+    struct SetReverseLookupEventsV1 has key, store {
+        set_reverse_lookup_events: event::EventHandle<SetReverseLookupEventV1>,
```
movekevin, line 79: Any reason for keeping this separate from ReverseLookupRegistryV1?

core/sources/domains.move
movekevin: Hmm I see.

movekevin (commented):
What happens to a reverse lookup mapping from address => domain if the domain expires or is forcefully revoked?

core/sources/domains.move
0xChucky: Separating events from their data is the pattern we went with, so I'm just sticking to that.

movekevin: I'm not sure if this is helpful or just spammy and inefficient storage wise (1 extra resource that contains 1 single event). Why not just add it to ReverseLookupRegistryV1? 🤔

core/Move.toml
movekevin: What were the errors you ran into?

core/sources/domains.move
0xChucky: Done

0xChucky:
> What happens to a reverse lookup mapping from address => domain if the domain expires or is forcefully revoked?

Good catch - in the forceful revoke path (i.e. `force_create_or_seize_domain_name()`), I fixed it to clear the reverse lookup for the target address if the target address's reverse lookup is the name being seized.

In the expiry case, it will remain the primary name on-chain until someone re-mints the name and sets it as their primary name. Expiration in general is enforced off-chain.

I also fixed `clear_name_address()` to clear the primary name if the name is a primary name.

core/Move.toml
0xChucky: Check out the github action presubmit from right before my commit which updates this file: https://github.com/aptos-labs/aptos-names-contracts/actions/runs/3705462363/jobs/6279398045

movekevin: That's strange 🤔 Can you try framework-mainnet instead of mainnet? This is the separate branch to track mainnet Move binaries.

core/sources/domains.move
movekevin: Anyway this is likely not a blocker.

core/sources/test_helper.move
```diff
@@ -134,14 +141,29 @@ module aptos_names::test_helper {
...
+            assert!(set_reverse_lookup_event_v1_num_emitted == 0, set_reverse_lookup_event_v1_num_emitted);
+        };
+
         if (is_subdomain) {
-            // We haven't set a target address yet!
-            test_utils::print_actual_expected(b"set_name_address_event_v1_num_emitted: ", set_name_address_event_v1_num_emitted, 0, false);
-            assert!(set_name_address_event_v1_num_emitted == 0, set_name_address_event_v1_num_emitted);
+            if (option::is_none(&user_reverse_lookup_before)) {
+                // Should automatically point to the users address
+                test_utils::print_actual_expected(b"set_name_address_event_v1_num_emitted: ", set_name_address_event_v1_num_emitted, 1, false);
```
movekevin, line 160: Should these print statements (print_actual_expected) be removed?

movekevin (approved):
Can you also add some more tests for the force seizing domain case?

core/sources/test_helper.move
0xChucky: I don't think so -- is there something here that makes it unclear?

0xChucky:
Added/updated some tests for seizing.

core/sources/domains.move
```diff
@@ -71,6 +71,19 @@ module aptos_names::domains {
         registry: Table<NameRecordKeyV1, NameRecordV1>,
     }
 
+    /// The registry for reverse lookups (aka primary names), which maps addresses to their primary name
+    /// - Users can set their primary name to a name that they own. This also forces the name to target their own address too.
+    /// - If you point your primary name to an address that isn't your own, it is no longer your primary name.
+    /// - If you mint a name while you don't have a primary name, your minted name is auto-set to be your primary name. (including minting subdomains)
+    struct ReverseLookupRegistryV1 has key, store {
```
geekflyer, line 78: I think a name like `ReverseLookupPrimaryNameRegistryV1` would a better name to indicate that this map can only be used to reverse lookup primary names, not secondary ones.

geekflyer:
I haven't done a detailed code review, but I would hightly suggest to use the term "Primary Name" instead of "Reverse Lookup Name" more actively in the code (name of structs and functions). E.g. `ReverseLookupRegistryV1 > PrimaryNameRegistryV1`.
As your branch name, pr description and our slack discussions indicate primary name is a term that people understand well and the name sort of implies that there can be only one.

"Reverse Lookup" on the other hand rather describes more of a process and from the name alone it's not clear that there can be only one name that can be reverse-looked up. In fact at some point we, and probably should be able to reverse-lookup all names of an address (primary and otherwise), although this would be probably difficult to achieve purely on-chain, but should be easy to achieve with the ANS db / an indexer off-chain.

movekevin (approved):

PR #6, opened 2023-01-10
[WIP] Adds other TLDs

0xChucky:


core/sources/utf8_utils.move
```diff
@@ -12,6 +12,36 @@ module aptos_names::utf8_utils {
...
+        let u64_str = utf8_to_vec_u64_fast(u8_str);
+        ends_with(&u64_str, suffix_bytes)
+    }
+
+    // Returns |true| if |str| ends with |suffix_bytes|.
+    public fun ends_with(str: &vector<u64>, suffix_bytes: &vector<u8>): bool {
+        let suffix = utf8_to_vec_u64_fast(suffix_bytes);
+        let suffix_len = vector::length(&suffix);
+        let str_len = vector::length(str);
+        if (suffix_len > str_len) {
```
0xchloe, line 28: Is there a reason that we are not using `>=`? as an example, I feel like maybe `.move` is not a viable domain name either?

core/sources/utf8_utils.move
```diff
@@ -21,6 +51,10 @@ module aptos_names::utf8_utils {
         let len = vector::length(&u64_characters);
         let allowed = true;
 
+        if (ends_with(&u64_characters, &MOVE_DOMAIN_SUFFIX)) {
```
0xchloe, line 54: what if people try to register a name like `charles.move.apt`?

banool: Perhaps the policy should be:
1. Dots are banned in names.
2. For convenience, if there is a single dot in the name and the suffix is in the valid set, remove the dot and suffix.
3. If there is more than 1 dot, abort.

0xChucky: if i understand the question correctly, `charles.move.apt` would abort because the loop below would see a `.` in the name and return false

core/sources/utf8_utils.move
0xChucky: ends_with should return true for calls like `ends_with(".xyz", ".xyz")` because ".xzy" ends with ".xyz"

core/sources/utf8_utils.move
```diff
@@ -12,6 +12,36 @@ module aptos_names::utf8_utils {
...
+    const MOVE_DOMAIN_SUFFIX: vector<u8> = b".move";
+
+    public fun ends_with_str(str: String, suffix_bytes: &vector<u8>): bool {
+        let u8_str = string::bytes(&str);
+        let u64_str = utf8_to_vec_u64_fast(u8_str);
+        ends_with(&u64_str, suffix_bytes)
+    }
+
+    // Returns |true| if |str| ends with |suffix_bytes|.
+    public fun ends_with(str: &vector<u64>, suffix_bytes: &vector<u8>): bool {
```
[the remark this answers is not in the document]
0xChucky: convenience, because we have vector<u64>/vector<u8> from where i'm currently calling it :)

PR #18, opened 2023-02-17
[Primary Name] merge testnet branch into mainnet branch 

0xZihan:
Merge testnet into mainnet so that mainnet has the latest changes for Primary Name.

### DIFF
--- a/core/sources/domains.move
+++ b/core/sources/domains.move
@@ -390,11 +390,11 @@ module aptos_names::domains {
     }
 
     /// Check if the address is the owner of the given aptos_name
-    /// If the name does not exist, returns false
-    public fun is_owner_of_name(owner_address: address, subdomain_name: Option<String>, domain_name: String): (bool, TokenId) {
+    /// If the name does not exist or owner owns an expired name, returns false
+    public fun is_owner_of_name(owner_address: address, subdomain_name: Option<String>, domain_name: String): (bool, TokenId) acquires NameRegistryV1 {
         let token_data_id = token_helper::build_tokendata_id(token_helper::get_token_signer_address(), subdomain_name, domain_name);
         let token_id = token_helper::latest_token_id(&token_data_id);
-        (token::balance_of(owner_address, token_id) > 0, token_id)
+        (token::balance_of(owner_address, token_id) > 0 && !name_is_expired(subdomain_name, domain_name), token_id)
     }
 
     /// gets the address pointed to by a given name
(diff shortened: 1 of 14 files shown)


core/sources/domains.move
```diff
@@ -135,7 +157,19 @@ module aptos_names::domains {
         token_helper::initialize(account);
     }
 
-    fun register_domain_generic(sign: &signer, domain_name: String, num_years: u8) acquires NameRegistryV1, RegisterNameEventsV1, SetNameAddressEventsV1 {
+    public fun init_reverse_lookup_registry_v1(account: &signer) {
```
movekevin, line 160: Not sure if I missed this in the initial review for testnet but can we change this to public entry fun? There's no reason for this to be non-entry and having to execute this using the script (sources/script.move). Changing from public fun to public entry fun is a compatible upgrade, so we should likely make a similar change in the testnet branch and upgrade the testnet contract.

core/sources/domains.move
```diff
@@ -135,7 +157,19 @@ module aptos_names::domains {
...
+        assert!(signer::address_of(account) == @aptos_names, error::permission_denied(ENOT_AUTHORIZED));
```
movekevin, line 161: For future: This assertion is technically not necessary if we don't plan to delay enabling primary domain after we do the contract upgrade. But let's keep this for now since we've already done this in testnet. For the future, let's not do this unless necessary.

init_reverse_lookup_registry_v1/sources/script.move
```diff
@@ -0,0 +1,7 @@
+script {
```
movekevin, line 1: This can be removed after updating init_reverse_lookup_registry_v1 to be public entry fun

movekevin (approved):

core/sources/domains.move
0xZihan: but to be safe should we add a check in `init_reverse_lookup_registry_v1` to make sure only creating the revers lookup table when there isn't one?

init_reverse_lookup_registry_v1/sources/script.move
0xZihan: I'll update it in the next version.

core/sources/domains.move
movekevin: Technically you don't need to. move_to will fail if you're creating the same resource twice for the same account. The error message is not as good as an explicit assertion but in some cases like this where a function is only meant to be called once, we don't need nice error messages.

PR #20, opened 2023-03-17
change addresses to use placeholders  on .toml file

0xmaayan:


banool (approved):

davidiw:
Why the change?

0xmaayan:
> Why the change?

* I want to use this repo when testing ANS on SDK (i.e clone this repo -> publish to local testnet -> test against it). If we have defaults addresses I can't publish it locally (talked to @0xZihan and @angieyth about it)
* looks like it is a convention to use `_` placeholder..?

PR #23, opened 2023-07-19
Aptos Name Service v2

BriungRi:


core_v2/sources/config.move
```diff
@@ -0,0 +1,498 @@
...
+        config: PropertyMap,
+    }
+
+    public(friend) fun initialize_v1(framework: &signer, admin_address: address, fund_destination_address: address) acquires ConfigurationV1 {
+        move_to(framework, ConfigurationV1 {
+            config: property_map::empty(),
+        });
+
+        // Temporarily set this to framework to allow other methods below to be set with framework signer
+        set_v1(@aptos_names_v2, config_key_admin_address(), &signer::address_of(framework));
```
davidiw, line 59: again these v1s are confusing a bit

core_v2/sources/config.move
```diff
@@ -0,0 +1,498 @@
...
+        set_v1(@aptos_names_v2, config_key_fund_destination_address(), &fund_destination_address);
+        set_v1(@aptos_names_v2, config_key_admin_address(), &admin_address);
+    }
+
+
+    //
+    // Configuration Shortcuts
+    //
+
+    public fun octas(): u64 {
```
davidiw, line 93: probably better to name this as apt_to_octa and take in a u64, no?

core_v2/sources/config.move
```diff
@@ -0,0 +1,498 @@
...
+
+    public fun read_bool_v1(addr: address, key: &String): bool acquires ConfigurationV1 {
+        property_map::read_bool(&borrow_global<ConfigurationV1>(addr).config, key)
+    }
+
+    public fun read_unvalidated_public_key(addr: address, key: &String): UnvalidatedPublicKey acquires ConfigurationV1 {
+        let value = property_map::borrow_value(property_map::borrow(&borrow_global<ConfigurationV1>(addr).config, key));
+        // remove the length of this vector recorded at index 0
+        vector::remove(&mut value, 0);
+        ed25519::new_unvalidated_public_key_from_bytes(value)
```
davidiw, line 350: this is weird, why did we encode the length here...

core_v2/sources/domain_e2e_tests.move
```diff
@@ -0,0 +1,890 @@
...
+    use aptos_names_v2::domains;
+    use aptos_names_v2::time_helper;
+    use aptos_names_v2::test_helper;
+    use aptos_names_v2::test_utils;
+    use std::option;
+    use std::signer;
+    use std::string;
+    use std::vector;
+
+    const MAX_REMAINING_TIME_FOR_RENEWAL_SEC: u64 = 15552000;
```
davidiw, line 16: might be good to comment how we came up with this seemingly arbitrary number

core_v2/sources/domain_e2e_tests.move
```diff
@@ -0,0 +1,890 @@
...
+
+    #[test(
```
davidiw, line 18: usually prefer we define the test execution at the top, that way when I read it, I challenge myself to equate the two...

core_v2/sources/domain_e2e_tests.move
```diff
@@ -0,0 +1,890 @@
...
+        router_signer = @router_signer,
+        aptos_names_v2 = @aptos_names_v2,
+        user = @0x077,
+        aptos = @0x1,
+        rando = @0x266f,
+        foundation = @0xf01d
+    )]
+    #[expected_failure(abort_code = 327682, location = aptos_names_v2::domains)]
```
davidiw, line 330: might be better to use hex here instead of decimal.

core_v2/sources/domain_e2e_tests.move
```diff
@@ -0,0 +1,890 @@
...
+        router_signer: &signer,
+        aptos_names_v2: &signer,
+        user: signer,
+        aptos: signer,
+        rando: signer,
+        foundation: signer,
+    ) {
+        let users = test_helper::e2e_test_setup(aptos_names_v2, user, &aptos, rando, &foundation);
+        let user = vector::borrow(&users, 0);
+        let rando = vector::borrow(&users, 1);
```
davidiw, line 369: I imagine these are just signers, why are we returning them like this and adding an extra 2-3 lines per test?

davidiw: maybe the test should take in a &signer?

core_v2/sources/price_model.move
```diff
@@ -0,0 +1,121 @@
...
+    use std::error;
+
+
+    /// The domain length is too short- currently the minimum is 2 characters
+    const EDOMAIN_TOO_SHORT: u64 = 1;
+    const SECONDS_PER_YEAR: u64 = 60 * 60 * 24 * 365;
+
+    /// The longer the name is registered for, the more expensive it is per year.
+    /// The curve is exponential, with every year costing more than the previous
+    fun scale_price_for_years(price: u64, years: u8): u64 {
```
davidiw, line 13: let's just delete this...

core_v2/sources/price_model.move
```diff
@@ -0,0 +1,121 @@
...
+        while (i <= years) {
+            final_price = final_price + (price * multiplier) / 100;
+            multiplier = multiplier + 5 + i * 5;
+            i = i + 1;
+        };
+        final_price
+    }
+
+    /// There is a fixed cost per each tier of domain names, from 2 to >=6, and it also scales exponentially with number of years to register
+    public fun price_for_domain_v1(domain_length: u64, registration_secs: u64): u64 {
```
davidiw, line 28: these `v1` are distracting lol

I would just make it a fixed price and kill all this code about dynamic prices....

core_v2/sources/price_model.move
```diff
@@ -0,0 +1,121 @@
...
+        assert!(price == 204, price);
+
+        let price = price_for_domain_v1(10, SECONDS_PER_YEAR) / config::octas();
+        assert!(price == 5, price);
+
+        let price = price_for_domain_v1(15, SECONDS_PER_YEAR) / config::octas();
+        assert!(price == 5, price);
+
+        let price = price_for_domain_v1(15, 10 * SECONDS_PER_YEAR) / config::octas();
+        assert!(price == 102, price);
```
davidiw, line 84: we shouldn't have to test for anything but 1 year beyond the current expiration or less out... and if someone wants to buy a lease for 6 months, they still pay the 1 year price :facepalm:

core_v2/sources/readme.md
```diff
@@ -0,0 +1,100 @@
+# Aptos Naming Service
```
davidiw, line 1: rofl, this is not near the top of the diff... so I read all that code up until now without any guidance...

core_v2/sources/readme.md
```diff
@@ -0,0 +1,100 @@
...
+  as `simplified_chinese`, or `emojis`, such as this unicorn: 🦄. Aptos is a unicorn-friendly zone. Neigh!
+
+### Character Sets
+
+With great UTF8 flexibility comes great UTF8 pain. In order to avoid this pain out of the gate, we are launching with a
+limited character set, which parallels ASCII English letters and numbers with the characters it supports. We are
+referring to
+this character set as `latin`.
+We will expand both the number of character sets we support, and the glyphs within each one, as quickly as possible to
+do safely.
```
davidiw, line 30: could we make a single line for a paragraph? it makes it easier to read in most text editors, github, and git does a good job of diffs on paragraphs now :raised_hands:

core_v2/sources/readme.md
```diff
@@ -0,0 +1,100 @@
...
+
+1. characters `a-z`, `0-9`, and `-` are allowed.
+2. Names may not begin or end with a `-`.
+3. Names may not contain more than 63 characters.
+
+### On-Chain Representation
+
+`Domains` and `Subdomains` are represented on-chain in two ways:
+
+1. As an `NFT`, following our Token Standard. When a name is registered, an NFT resource is created and `moved` into the
```
davidiw, line 47: probably good to call this a digital asset and point to aip-11 https://github.com/aptos-foundation/AIPs/edit/main/aips/aip-11.md

davidiw (commented):
I have a lot more to read, but this is my first set of notes... I think we should strip out functionality from v2 that we don't need:
* Signatures
* Scaling of prices

Would also suggest we eliminate these `v1` or versioned APIs. If we keep the internal API as `public(friend)`, then we can upgrade it without issue.

core_v2/sources/domains.move
```diff
@@ -0,0 +1,1232 @@
...
+    /// The subdomain expiration policy is included in the enum SUBDOMAIN_POLICY_*
+    const ESUBDOMAIN_EXPIRATION_POLICY_INVALID: u64 = 26;
+    /// Caller must be the router
+    const ENOT_ROUTER: u64 = 27;
+
+    /// Tokens require a signer to create, so this is the signer for the collection
+    struct CollectionCapability has key, drop {
+        capability: SignerCapability,
+        burn_signer_capability: SignerCapability,
+    }
```
davidiw, line 82: Do we really want to continue to leverage this legacy method for managing the service?

core_v2/sources/domains.move
```diff
@@ -0,0 +1,1232 @@
...
+
+    /// Manager object refs
+    struct Manager has key {
```
davidiw, line 85: why isn't this a resource group?

core_v2/sources/domains.move
```diff
@@ -0,0 +1,1232 @@
...
+        /// The extend_ref of the manager object to get its signer
+        extend_ref: object::ExtendRef,
+    }
+
+    struct SubdomainExt has store {
```
davidiw, line 90: please add comments around what this and how it is used.

core_v2/sources/domains.move
```diff
@@ -0,0 +1,1232 @@
...
+        subdomain_name: String,
+        subdomain_expiration_policy: u8,
+    }
+
+    #[resource_group_member(group = aptos_framework::object::ObjectGroup)]
+    struct NameRecord has key {
```
davidiw, line 96: Usually I try to put structs in order of first appearance and so this should be ahead of `SubdomainExt`

core_v2/sources/domains.move
```diff
@@ -0,0 +1,1232 @@
...
+
+    /// Holder for `RegisterNameEvent` events
+    struct RegisterNameEvents has key, store {
+        register_name_events: event::EventHandle<RegisterNameEvent>,
+    }
+
+    /// Holder for `RenewNameEvent` events
+    struct RenewNameEvents has key, store {
+        renew_name_events: event::EventHandle<RenewNameEvent>,
+    }
```
davidiw, line 129: why aren't these in the object group and a single resource?

core_v2/sources/domains.move
```diff
@@ -0,0 +1,1232 @@
...
+            APP_SIGNER_CAPABILITY_SEED,
+        );
+        let (_, burn_signer_capability) = account::create_resource_account(
+            account,
+            BURN_SIGNER_CAPABILITY_SEED,
+        );
+        move_to(account, CollectionCapability {
+            capability: token_signer_cap,
+            burn_signer_capability,
+        });
```
davidiw, line 206: do you have a pointer to the rationale for the security / authorization model?

core_v2/sources/domains.move
```diff
@@ -0,0 +1,1232 @@
...
+    }
+
+    public entry fun force_set_name_expiration(
+        sign: &signer,
+        domain_name: String,
+        subdomain_name: Option<String>,
+        new_expiration_secs: u64
+    ) acquires CollectionCapability, NameRecord {
+        // check the signer eligibility
+        config::assert_signer_is_admin(sign);
```
davidiw, line 630: let's call this admin instead of sign

core_v2/sources/domains.move
```diff
@@ -0,0 +1,1232 @@
...
+        expiration_time_secs: u64,
+
+        // Extras for indexing
+        target_address: Option<address>,
+        is_primary_name: bool,
+    }
+
+    /// This is only callable during publishing
+    fun init_module(account: &signer) {
+        let funds_address: address = @aptos_names_funds;
```
movekevin, line 175: nit: The :address part is not necessary

core_v2/sources/domains.move
```diff
@@ -0,0 +1,1232 @@
...
+        } else {
+            config::domain_collection_name()
+        }
+    }
+
+    fun validate_registration_duration(
+        registration_duration_secs: u64,
+    ) {
+        assert!(
+            registration_duration_secs % SECONDS_PER_YEAR == 0,
```
movekevin, line 998: Why even take registration_duration_secs as seconds then? Why not take the number of years and convert to seconds here?

core/sources/domains.move
```diff
@@ -427,22 +427,43 @@ module aptos_names::domains {
         if (option::is_some(&subdomain_name) && name_is_registerable(option::none(), domain_name)) {
             return false
         };
-        // Check to see if the domain is registered, or expired
-        let aptos_names = borrow_global<NameRegistryV1>(@aptos_names);
```
movekevin, line 431: This PR is massive already. Can we split out changes to v1 into a different PR?

core_v2/sources/v2_price_model.move
```diff
@@ -0,0 +1,121 @@
...
+    use std::error;
+
+
+    /// The domain length is too short- currently the minimum is 2 characters
+    const EDOMAIN_TOO_SHORT: u64 = 1;
+    const SECONDS_PER_YEAR: u64 = 60 * 60 * 24 * 365;
+
+    /// The longer the name is registered for, the more expensive it is per year.
+    /// The curve is exponential, with every year costing more than the previous
+    fun scale_price_for_years(price: u64, years: u8): u64 {
```
davidiw, line 13: again, why did we keep this code, we are only support 1 year buy outs for a fixed rate... can we please remove this.

core_v2/sources/v2_price_model.move
```diff
@@ -0,0 +1,121 @@
...
+        let i = 1;
+        while (i <= years) {
+            final_price = final_price + (price * multiplier) / 100;
+            multiplier = multiplier + 5 + i * 5;
+            i = i + 1;
+        };
+        final_price
+    }
+
+    /// There is a fixed cost per each tier of domain names, from 2 to >=6, and it also scales exponentially with number of years to register
```
davidiw, line 27: this isn't even defined or part of any decision, we shouldn't keep this type of code

core_v2/sources/v2_price_model.move
```diff
@@ -0,0 +1,121 @@
...
+        let length_to_charge_for = math64::min(domain_length, 6);
+        let registration_years = (registration_secs / SECONDS_PER_YEAR as u8);
+        scale_price_for_years(v2_config::domain_price_for_length(length_to_charge_for), registration_years)
+    }
+
+    /// Subdomains have a fixed unit cost
+    public fun price_for_subdomain(_registration_duration_secs: u64): u64 {
+        v2_config::subdomain_price()
+    }
+
```
davidiw, line 39: I think we should just eliminate this as well since subdomains will be free from here on out...

core_v2/sources/v2_token_helper.move
```diff
@@ -0,0 +1,65 @@
...
+
+    const DOMAIN_DELIMITER: vector<u8> = b".";
+    const DOMAIN_SUFFIX: vector<u8> = b".apt";
+
+    /// The collection does not exist. This should never happen.
+    const ECOLLECTION_NOT_EXISTS: u64 = 1;
+
+    public fun get_fully_qualified_domain_name(subdomain_name: Option<String>, domain_name: String): String {
+        let (domain_is_allowed, _length) = v2_utf8_utils::string_is_allowed(&domain_name);
+        assert!(domain_is_allowed, 1);
```
davidiw, line 16: we should use proper errors.

core_v2/sources/v2_token_helper.move
```diff
@@ -0,0 +1,65 @@
...
+        let subdomain_is_allowed = if (option::is_some(&subdomain_name)) {
+            let (subdomain_is_allowed, _length) = v2_utf8_utils::string_is_allowed(option::borrow(&subdomain_name));
+            subdomain_is_allowed
+        } else {
+            true
+        };
+        assert!(subdomain_is_allowed, 2);
```
davidiw, line 23: proper errors

core_v2/sources/v2_token_helper.move
```diff
@@ -0,0 +1,65 @@
...
+    use std::option::{Self, Option};
+    use std::string::{Self, String};
+
+    const DOMAIN_DELIMITER: vector<u8> = b".";
+    const DOMAIN_SUFFIX: vector<u8> = b".apt";
+
+    /// The collection does not exist. This should never happen.
+    const ECOLLECTION_NOT_EXISTS: u64 = 1;
+
+    public fun get_fully_qualified_domain_name(subdomain_name: Option<String>, domain_name: String): String {
```
davidiw, line 14: I don't see why we need this separate from v2_domains.move

core_v2/sources/v2_time_helper.move
```diff
@@ -0,0 +1,52 @@
+module aptos_names_v2::v2_time_helper {
```
davidiw, line 1: I don't see why we need this separate from v2_domains.move

core_v2/sources/query_helper.move
```diff
@@ -0,0 +1,109 @@
+module aptos_names_v2::query_helper {
```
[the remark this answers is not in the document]
davidiw: if this were in the router it would be fine, but we didn't migrate it there... can we please merge this into v2_domains?

core_v2/sources/v2_utf8_utils.move
```diff
@@ -0,0 +1,193 @@
+module aptos_names_v2::v2_utf8_utils {
```
davidiw, line 1: probably better to call this domain_validator or string_validator

core_v2/sources/v2_domains.move
```diff
@@ -0,0 +1,1225 @@
...
+
+        validate_registration_duration(registration_duration_secs);
+
+        let subdomain_name = option::none<String>();
+
+        assert!(is_name_registerable(domain_name, subdomain_name), error::invalid_state(ENAME_NOT_AVAILABLE));
+
+        let length = validate_name_string(domain_name);
+
+        let price = v2_price_model::price_for_domain(length, registration_duration_secs);
```
davidiw, line 294: considering what we discussed, I think this needs to be a flat fee and that we likely bump `registration_duration_secs` to the nearest next chunk of time`

I would be unhappy to pay full price and not get the full re-registration time allotted to me.

core_v2/sources/v2_domains.move
```diff
@@ -0,0 +1,1225 @@
...
+    /// A wrapper around `register_name` as an entry function.
+    /// Option<String> is not currently serializable, so we have these convenience methods
+    public fun register_domain(
+        router_signer: &signer,
+        sign: &signer,
+        domain_name: String,
+        registration_duration_secs: u64,
+    ) acquires CollectionCapability, NameRecord, RegisterNameEvents, ReverseRecord, SetReverseLookupEvents {
+        assert!(address_of(router_signer) == @router_signer, error::permission_denied(ENOT_ROUTER));
+        assert!(v2_config::unrestricted_mint_enabled(), error::permission_denied(EVALID_SIGNATURE_REQUIRED));
```
davidiw, line 284: why is this even here?

core_v2/sources/v2_domains.move
```diff
@@ -0,0 +1,1225 @@
...
+        set_name_events: event::EventHandle<SetTargetAddressEvent>,
+    }
+
+    /// Holder for `RegisterNameEvent` events
+    struct RegisterNameEvents has key, store {
+        register_name_events: event::EventHandle<RegisterNameEvent>,
+    }
+
+    /// Holder for `RenewNameEvent` events
+    struct RenewNameEvents has key, store {
```
davidiw, line 119: why are these key and store?

also put them in a common resource group

also if we still have time, it might be more prudent to make these eventv2

if not we can upgrade in the near future...

core_v2/sources/v2_domains.move
```diff
@@ -0,0 +1,1225 @@
...
+        let token_signer = object::generate_signer(&constructor_ref);
+        // creating subdomain
+        if (is_subdomain(subdomain_name)) {
+            subdomain_ext = option::some(SubdomainExt {
+                subdomain_name: option::extract(&mut subdomain_name),
+                subdomain_expiration_policy: SUBDOMAIN_POLICY_MANUAL_SET_EXPIRATION,
+            });
+        }
+        // creating domain
+        else {
```
davidiw, line 257:
```suggestion
        } else { // creating domain
```
this is a bit unnatural syntax

core_v2/sources/v2_domains.move
```diff
@@ -0,0 +1,1225 @@
...
+    }
+
+    /// Register a name. Accepts an optional subdomain name, a required domain name, and a registration duration in seconds.
+    /// For domains, the registration duration is only allowed to be in increments of 1 year, for now
+    /// Since the owner of the domain is the only one that can create the subdomain, we allow them to decide how long they want the underlying registration to be
+    /// The maximum subdomain registration duration is limited to the duration of its parent domain registration
+    ///
+    /// NOTE: Registration validation already done. This function does not perform any validation on whether `sign` is allowed to register this name
+    fun register_name_internal(
+        sign: &signer,
```
davidiw, line 365: can we label this instead to buyer or purchaser or creator? `sign` is not a very descriptive name for the actor here.

core_v2/sources/v2_domains.move
```diff
@@ -0,0 +1,1225 @@
...
+                is_primary_name,
+            },
+        );
+    }
+
+    // === SUBDOMAIN MANAGEMENT ===
+
+    /// Disable or enable subdomain owner from transferring subdomain as domain owner
+    public fun set_subdomain_transferability_as_domain_owner(
+        router_signer: &signer,
```
davidiw, line 474: I think we can just call this `router` as there's `router_signer` doesn't add more clarity, but that's more a nit

core_v2/sources/v2_domains.move
```diff
@@ -0,0 +1,1225 @@
...
+        let name_record_address = get_token_addr(domain_name, option::some(subdomain_name));
+        let transfer_ref = &borrow_global_mut<NameRecord>(name_record_address).transfer_ref;
+        if (transferrable) {
+            object::enable_ungated_transfer(transfer_ref);
+        } else {
+            object::disable_ungated_transfer(transfer_ref);
+        }
+    }
+
+    public fun transfer_subdomain_owner(
```
davidiw, line 491: I don't think we need this and it isn't being used in this contract, let's remove it. if we ever get some dispatch service in move, this will be useful.

bulk_migrate/sources/migrate.move
```diff
@@ -0,0 +1,28 @@
...
+        );
+        let idx = 0;
+        while (idx < vector::length(&domain_names)) {
+            let domain_name = *vector::borrow(&mut domain_names, idx);
+            let subdomain_name = *vector::borrow(&mut subdomain_names, idx);
+            router::migrate_name(user, domain_name, subdomain_name);
+            idx = idx + 1
+        }
+    }
+}
```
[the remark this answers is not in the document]
davidiw: As a minimalist, I tend to prefer we only produce interfaces that we intend or must support. Maybe this is within the scope, but it feels extra.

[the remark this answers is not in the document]
davidiw: I understand that, but what website is going to offer this functionality?

core/sources/config.move
```diff
@@ -95,10 +95,16 @@ module aptos_names::config {
         100000000
     }
 
+    // !!!!! DEPRECATED !!!!! please use is_enabled_for_nonadmin()
     public fun is_enabled(): bool acquires ConfigurationV1 {
         read_bool_v1(@aptos_names, &config_key_enabled())
     }
 
+    // Returns true if enabled flag is true or signer is admin
+    public fun is_enabled_for_nonadmin(sign: &signer): bool acquires ConfigurationV1 {
```
movekevin, line 104: nit: Why does this need to take a signer if it doesn't need to do anything with it? Why not take an address instead?

core_v2/sources/v2_config.move
```diff
@@ -0,0 +1,476 @@
...
+    const EINVALID_VALUE: u64 = 2;
+
+    const SECONDS_PER_YEAR: u64 = 60 * 60 * 24 * 365;
+
+    struct Config has key, store {
+        config: PropertyMap,
+    }
+
+    public(friend) fun initialize_config(
+        framework: &signer,
```
movekevin, line 52: nit: This is a bit misleading as framework signer often means 0x1 signer but this is not it

router/sources/router.move
```diff
@@ -0,0 +1,928 @@
+module router::router {
```
movekevin, line 1: Can you add more documentations on when these different functions are used? There's some overlapping in a few of them such as migrate and renew

router/sources/router.move
```diff
@@ -0,0 +1,928 @@
...
+        } else if (mode == MODE_V1_AND_V2) {
+            assert!(can_register_in_v2(domain_name, option::none()), error::unavailable(ENAME_NOT_AVAILABLE));
+            v2_domains::register_domain(
+                &get_router_signer(),
+                user,
+                domain_name,
+                registration_duration_secs,
+            );
+            // Clear the name in v1
+            domains::force_clear_registration(&get_router_signer(), option::none(), domain_name)
```
movekevin, line 197: This doesn't clear the subdomains right? Would this be an issue?

router/sources/router.move
```diff
@@ -0,0 +1,928 @@
...
+    /// @param user The user who is paying for the registration
+    /// @param domain_name The domain name to register
+    /// @param registration_duration_secs The duration of the registration in seconds
+    /// @param target_addr The address the registered name will point to
+    /// @param to_addr The address to send the token to. If none, then the user will be the owner. In MODE_V1, receiver must have already opted in to direct_transfer via token::opt_in_direct_transfer
+    public entry fun register_domain(
+        user: &signer,
+        domain_name: String,
+        registration_duration_secs: u64,
+        target_addr: Option<address>,
```
movekevin, line 174: Is it possible to simplify here and always require users to specify target and to addresses? They can also way set to user's address initially and change it later if needed. This would greatly simplify the code + reduce branches. Also makes the function easier to call not having to deal with options

router/sources/router.move
```diff
@@ -0,0 +1,928 @@
...
+        if (option::is_some(&to_addr) && to_addr != option::some(owner_addr)) {
+            return
+        };
+
+        // if the target address is not the buyer address
+        if (option::is_some(&target_addr) && target_addr != option::some(owner_addr)) {
+            return
+        };
+
+        let has_primary_name = has_primary_name(user);
```
movekevin, line 243: nit: Combine this with the next line if (!has_primary_name(user)) {...}

router/sources/router.move
```diff
@@ -0,0 +1,928 @@
...
+            v2_domains::is_name_registerable(domain_name, subdomain_name)
+        }
+    }
+
+    /// @param user The user who is paying for the registration
+    /// @param domain_name The domain name to register
+    /// @param registration_duration_secs The duration of the registration in seconds
+    /// @param target_addr The address the registered name will point to
+    /// @param to_addr The address to send the token to. If none, then the user will be the owner. In MODE_V1, receiver must have already opted in to direct_transfer via token::opt_in_direct_transfer
+    public entry fun register_domain(
```
movekevin, line 170: Can we document the different behaviors for v1 vs v2?

router/sources/router.move
```diff
@@ -0,0 +1,928 @@
...
+    // ==== REGISTRATION ====
+
+    /// If the name is registered and active in v1, then the name can only be registered if we have burned the token (sent it to the router_signer)
+    /// Else, the name can only be registered if it is available in v2 (we double check availablity for safety)
+    inline fun can_register_in_v2(domain_name: String, subdomain_name: Option<String>): bool {
+        if (domains::name_is_registered(
+            subdomain_name,
+            domain_name
+        ) && !domains::name_is_expired(subdomain_name, domain_name)) {
+            let (is_burned, _token_id) = domains::is_token_owner(
```
movekevin, line 154: nit: You can just do _ instead of _token_id

router/sources/router.move
```diff
@@ -0,0 +1,928 @@
...
+        };
+
+        // if the target address is not the buyer address
+        if (option::is_some(&target_addr) && target_addr != option::some(owner_addr)) {
+            return
+        };
+
+        let has_primary_name = has_primary_name(user);
+        if (!has_primary_name) {
+            set_primary_name(user, domain_name, subdomain_name);
```
movekevin, line 245: Why does set_primary_name also migrate the domain as well instead of just setting the primary name? This also means register_domain also does the migration at the same time, which is overlapping with migrate_name. Why not just disallow that and only allow migrate_name to do the migration to avoid extra complexity and branches? One function one task = much easier to reason, test and debug

router/sources/router.move
```diff
@@ -0,0 +1,928 @@
...
+                domain_name,
+                option::some(subdomain_name),
+                *option::borrow(&target_addr)
+            );
+        };
+        if (option::is_some(&to_addr)) {
+            transfer_name(user, domain_name, option::some(subdomain_name), *option::borrow(&to_addr));
+        };
+        if (mode == MODE_V1_AND_V2) {
+            v2_domains::set_subdomain_transferability_as_domain_owner(
```
movekevin, line 306: Why not do this in the branch on line 270?

router/sources/router.move
```diff
@@ -0,0 +1,928 @@
...
+            // Get the v1 token info
+            let (
+                _property_version,
+                expiration_time_sec,
+                target_addr
+            ) = domains::get_name_record_v1_props_for_name(
+                subdomain_name,
+                domain_name,
+            );
+            let tokendata_id = aptos_names::token_helper::build_tokendata_id(
```
movekevin, line 373: Is the full namespace here required because of naming conflicts?

router/sources/router.move
```diff
@@ -0,0 +1,928 @@
...
+                aptos_names::token_helper::get_token_signer_address(),
+                subdomain_name,
+                domain_name,
+            );
+            let token_id = aptos_names::token_helper::latest_token_id(&tokendata_id);
+
+            // Domain must migrate before subdomain, throw error if this is a subdomain but domain has not been migrated
+            if (option::is_some(&subdomain_name)) {
+                assert!(
```
movekevin, line 382: Do we actually need this check? Users can't register subdomains without the domain existing right?

router/sources/router.move
```diff
@@ -0,0 +1,928 @@
...
+                token_id,
+                1,
+            );
+
+            // Calculate new expiration
+            let now = timestamp::now_seconds();
+            let new_expiration_time_sec = if (option::is_some(&subdomain_name)) {
+                expiration_time_sec
+            } else {
+                if (expiration_time_sec <= AUTO_RENEWAL_EXPIRATION_CUTOFF_SEC) {
```
movekevin, line 402: Can you add more documentation on the rule here? It looks like we don't want the migrated domain to have expiration < 6 months

core_v2/sources/v2_domains.move
```diff
@@ -0,0 +1,1318 @@
...
+
+    // === RENEW DOMAIN ===
+
+    public fun renew_domain(
+        sign: &signer,
+        domain_name: String,
+        renewal_duration_secs: u64,
+    ) acquires CollectionCapability, NameRecord, SubdomainExt, RenewNameEvents, ReverseRecord {
+        // check the domain eligibility
+        let length = validate_name_string(domain_name);
```
movekevin, line 430: this doesn't seem necessary? This should already been checked as part of registering the domain the first time

core_v2/sources/v2_domains.move
```diff
@@ -0,0 +1,1318 @@
...
+
+        validate_registration_duration(renewal_duration_secs);
+        assert!(is_domain_in_renewal_window(domain_name), error::invalid_state(EDOMAIN_NOT_AVAILABLE_TO_RENEW));
+        let price = v2_price_model::price_for_domain(length, renewal_duration_secs);
+        // pay the price
+        coin::transfer<AptosCoin>(sign, v2_config::fund_destination_address(), price);
+        renew_domain_internal(domain_name, renewal_duration_secs, price);
+    }
+
+    fun renew_domain_internal(
```
movekevin, line 440: Does this need to be a separate function? Looks like it's only used once

core_v2/sources/v2_domains.move
```diff
@@ -0,0 +1,1318 @@
...
+
+        // Idea here is that if this is a primary name, then the target_addr's reverse lookup should point back to this domain
+        let is_primary_name = if (option::is_some(&record.target_address)) {
+            let maybe_reverse_record = if (exists<ReverseRecord>(*option::borrow(&record.target_address))) {
+                let reverse_record = borrow_global<ReverseRecord>(*option::borrow(&record.target_address));
+                reverse_record.token_addr
+            } else {
+                option::none()
+            };
+            if (option::is_some(&maybe_reverse_record)) {
```
movekevin, line 456: Why not combine this if with the previous one (if (exists<ReverseRecord>(*option::borrow(&record.target_address))))? Seems unnecessary to break this into 2 separate conditionals and returning the option in between if none of the intermediate results is used outside

core_v2/sources/v2_domains.move
```diff
@@ -0,0 +1,1318 @@
...
+        let token_addr = get_token_addr_inline(domain_name, option::some(subdomain_name));
+        let record = borrow_global_mut<NameRecord>(token_addr);
+        record.target_address = new_target_address;
+        object::transfer_with_ref(object::generate_linear_transfer_ref(&record.transfer_ref), new_owner_address);
+        // clear the primary name
+        clear_reverse_lookup_for_name(option::some(subdomain_name), domain_name);
+    }
+
+    /// this is for domain owner to update subdomain expiration time
+    public fun set_subdomain_expiration(
```
movekevin, line 529: This is a bit of strange experience. If I want to change my subdomain from auto renew to having a custom expiration, I would need to do two steps: 
1. Call set_subdomain_expiration_policy to change the policy
2. Call set_subdomain_expiration to change the expiration to custom
Seems a bit complex

core_v2/sources/v2_domains.move
```diff
@@ -0,0 +1,1318 @@
...
+        );
+
+        assert!(
+            registration_duration_secs <= v2_config::max_number_of_seconds_registered(),
+            error::out_of_range(EINVALID_NUMBER_YEARS)
+        );
+    }
+
+
+    fun validate_subdomain_expiration_policy(
```
movekevin, line 1002: Do we plan to allow more types of policies? I actually can't think of any other possible policy beside these 2 so would it be simpler to just have a single auto_renew bool instead?

core_v2/sources/v2_domains.move
```diff
@@ -0,0 +1,1318 @@
...
+    /// For domains, the registration duration is only allowed to be in increments of 1 year, for now
+    /// Since the owner of the domain is the only one that can create the subdomain, we allow them to decide how long they want the underlying registration to be
+    /// The maximum subdomain registration duration is limited to the duration of its parent domain registration
+    ///
+    /// NOTE: Registration validation already done. This function does not perform any validation on whether `sign` is allowed to register this name
+    fun register_name_internal(
+        sign: &signer,
+        subdomain_name: Option<String>,
+        domain_name: String,
+        registration_duration_secs: u64,
```
movekevin, line 373: Is auto_renew = false (custom expiration) the default behavior for subdomain registration? If so, there could be a high chance no one uses auto-renew. Why not default to auto-renew?

core/sources/config.move
movekevin: This is kinda a view function anyway that has no side effect. The callers of this function can enforce that signer themselves.

router/sources/router.move
movekevin: This does create a lot of edge cases in all of the interactions though. It'd be strange for a user, for example, to want to change the primary name and not expecting anything but that. But to their surprise, the domain NFT disappears from their wallet (and a new one randomly appears). Is there a good solution for this on the FE side?

router/sources/router.move
movekevin: Is there not a check in the registration step already?

core_v2/sources/v2_domains.move
movekevin: OK

core_v2/sources/v2_domains.move
movekevin: Can't it be:

if (exists<ReverseRecord>(*option::borrow(&record.target_address))) {
     let reverse_record = borrow_global<ReverseRecord>(*option::borrow(&record.target_address));
     get_token_addr_inline(domain_name, option::none()) == reverse_record.token_addr
} else {
  false
}


?

core_v2/sources/v2_config.move
```diff
@@ -0,0 +1,388 @@
...
+        move_to(deployer, Config {
+            enabled: true,
+            admin_address,
+            fund_destination_address,
+            max_number_of_seconds_registered: SECONDS_PER_YEAR * 2,
+            max_domain_length: 63,
+            min_domain_length: 3,
+            tokendata_description: string::utf8(b"This is an official Aptos Labs Name Service Name"),
+            tokendata_url_prefix: string::utf8(b"https://www.aptosnames.com/api/mainnet/v1/metadata/"),
+            domain_price_length_3: 20 * octas(),
```
davidiw, line 65: these are not the current values for v1, please update to those.

core_v2/sources/v2_config.move
```diff
@@ -0,0 +1,388 @@
...
+    }
+
+    public entry fun set_tokendata_url_prefix(sign: &signer, url_prefix: String) acquires Config {
+        assert_signer_is_admin(sign);
+        borrow_global_mut<Config>(@aptos_names_v2).tokendata_url_prefix = url_prefix
+    }
+
+    public entry fun set_subdomain_price(sign: &signer, price: u64) acquires Config {
+        assert_signer_is_admin(sign);
+        borrow_global_mut<Config>(@aptos_names_v2).subdomain_price = price
```
davidiw, line 220: in the future you can further reduce this with a inline

```
inline fun authorized_borrow_mut(sign): &mut Config acquires Config {
  assert_signer_is_admin(sign);
  borrow_global_mut<Config>(@aptos_names_v2)
}
```

then you can reduce each of these by 50%

davidiw (commented):
will update with comments file by file.

core_v2/sources/v2_domains.move
```diff
@@ -0,0 +1,1315 @@
...
+    }
+
+    #[resource_group_member(group = aptos_names_v2::v2_domains::ObjectGroup)]
+    /// This is a subdomain extension that is only used for subdomains
+    struct SubdomainExt has key {
+        subdomain_name: String,
+        subdomain_expiration_policy: u8,
+    }
+
+    struct ReverseRecord has key {
```
davidiw, line 109: We probably should have had this in its own resource group so we can upgrade it in the future, if we deem it necessary. Is it too late?

core_v2/sources/v2_config.move
```diff
@@ -0,0 +1,388 @@
...
+    const ENOT_AUTHORIZED: u64 = 1;
+    /// Raised if there is an invalid value for a configuration
+    const EINVALID_VALUE: u64 = 2;
+    /// Domain length is invalid, it must be at least 3 characters
+    const EINVALID_DOMAIN_LENGTH: u64 = 3;
+
+    const SECONDS_PER_YEAR: u64 = 60 * 60 * 24 * 365;
+    const SECONDS_PER_DAY: u64 = 60 * 60 * 24;
+
+    #[resource_group_member(group = aptos_framework::object::ObjectGroup)]
```
davidiw, line 32: this probably should have been its own resource group, but I guess this work lol

core_v2/sources/v2_domains.move
```diff
@@ -0,0 +1,1315 @@
...
+            set_reverse_lookup_events: account::new_event_handle<SetReverseLookupEvent>(app_signer),
+        });
+        move_to(app_signer, DomainObject {
+            extend_ref,
+        });
+    }
+
+    /// Creates a token for the name.
+    /// NOTE: This function performs no validation checks
+    fun create_token(
```
davidiw, line 241: feels like create_token should not have any logic for subdomain and we should have a function that wraps this for create_domain and create_subdomain with whatever additional logic for example for subdomains we should have ... let's see what happens in the rest of the code.

core_v2/sources/v2_domains.move
```diff
@@ -0,0 +1,1315 @@
...
+        let price = v2_price_model::price_for_domain(length, registration_duration_secs);
+        coin::transfer<AptosCoin>(sign, v2_config::fund_destination_address(), price);
+
+        register_name_internal(sign, subdomain_name, domain_name, registration_duration_secs, price);
+    }
+
+    /// A wrapper around `register_name` as an entry function.
+    /// Option<String> is not currently serializable, so we have these convenience method
+    /// `expiration_time_sec` is the timestamp, in seconds, when the name expires
+    public fun register_subdomain(
```
davidiw, line 311: how do we the subdomain policies?

core_v2/sources/v2_domains.move
davidiw: I'll push back and argue that the behavior from v1 is NOT the right behavior. It is hard to manage and awkward.

core_v2/sources/v2_domains.move
davidiw: yeah, I agree with Kevin this isn't ... that complex of a nest

core_v2/sources/v2_domains.move
```diff
@@ -0,0 +1,1315 @@
...
+        // validate user own the domain
+        let signer_addr = signer::address_of(sign);
+        assert!(
+            is_token_owner(signer_addr, domain_name, option::none()),
+            error::permission_denied(ENOT_OWNER_OF_DOMAIN)
+        );
+        assert!(
+            !is_name_expired(domain_name, option::none()),
+            error::permission_denied(ECANNOT_TRANSFER_SUBDOMAIN_WHILE_DOMAIN_HAS_EXPIRED)
+        );
```
davidiw, line 515: Why does this matter?

core_v2/sources/v2_domains.move
```diff
@@ -0,0 +1,1315 @@
...
+        assert!(
+            subdomain_ext.subdomain_expiration_policy != SUBDOMAIN_POLICY_LOOKUP_DOMAIN_EXPIRATION,
+            error::invalid_state(ESUBDOMAIN_IS_AUTO_RENEW)
+        );
+
+        // manually set the expiration date
+        record.expiration_time_sec = expiration_time_sec;
+    }
+
+    public fun set_subdomain_expiration_policy(
```
davidiw, line 555: feels like this isn't the right API

we should probably just have distinct `set_subdomain_expiration_policy_as_xxx` functions

core_v2/sources/v2_domains.move
davidiw: fair point, Kevin, but it isn't really auto-renew, it is a distinct policy...

core_v2/sources/v2_domains.move
```diff
@@ -0,0 +1,1315 @@
...
+
+        let expiration_time_sec = get_expiration(domain_name, subdomain_name);
+
+        // Name is expired and passed grace period, so name is registerable
+        if (timestamp::now_seconds() > v2_config::reregistration_grace_sec() + expiration_time_sec) {
+            return true
+        } else {
+            // Name is expired but haven't passed grace period, so name is not registerable
+            return false
+        }
```
davidiw, line 1065: so ... we should eliminate this if and just return line 1060.

movekevin (approved):

PR #29, opened 2023-07-21
Add function to migrate from token v1 to token v2

BriungRi:
### Description

- Add an entry function to burn a user's token v1 and mint them a token v2 for the same name with a one year extension on the name
  - Burn works by sending the token to a `burn_signer` which is a resource account initialized by `aptos_names_v2::domains`
- Move `aptos_names_v2` to a different address from `aptos_names`
  - Update `signature` in `e2e_test_with_valid_signature` because signature involves the contract address (which has now changed)
- Extend the user's expiration by a year if the name expires before March 7, 2024
- TODOs for migrating primary name and for `register_name` to accept target_addr

### Test Plan

- New unit tests for migration, 1 with auto renewal, 1 without autorenewal

### DIFF
--- a/core_v2/sources/domain_e2e_tests.move
+++ b/core_v2/sources/domain_e2e_tests.move
@@ -477,7 +477,7 @@ module aptos_names_v2::domain_e2e_tests {
         rando = @0x266f,
         foundation = @0xf01d
     )]
-    fun test_migration(
+    fun test_migration_auto_renewal(
         aptos_names: &signer,
         aptos_names_v2: &signer,
         user: signer,
@@ -538,7 +538,7 @@ module aptos_names_v2::domain_e2e_tests {
             assert!(option::is_none(&target_addr), 3);
         };
 
-        // The v2 name belongs to `user`, the target address is the user_addr, and exipire in 1 + 1 years
+        // The v2 name belongs to `user`, the target address is the user_addr, and expire in 1 + 1 years
         {
             assert!(
                 aptos_names_v2::domains::name_is_registerable(option::none(), test_helper::domain_name()) == false,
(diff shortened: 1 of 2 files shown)


core_v2/sources/domain_e2e_tests.move
```diff
@@ -467,4 +468,96 @@ module aptos_names_v2::domain_e2e_tests {
...
+
+    #[test(
+        aptos_names = @aptos_names,
+        aptos_names_v2 = @aptos_names_v2,
+        user = @0x077,
+        aptos = @0x1,
+        rando = @0x266f,
+        foundation = @0xf01d
+    )]
+    fun test_migration(
```
angieyth, line 480: very minor thing. I noticed that the old tests are all names `**_test()` and I do like the convention with `test_*` more. should we unify everything here?

core_v2/sources/domains.move
```diff
@@ -749,6 +770,33 @@ module aptos_names_v2::domains {
...
+    ) acquires CollectionCapabilityV2, NameRecordV2, RegisterNameEventsV1, ReverseRecord, SetNameAddressEventsV1, SetReverseLookupEventsV1 {
+        let (expiration_time_sec, target_addr) = migrate_helper::burn_token_v1(
+            user,
+            &get_burn_signer(),
+            domain_name,
+            option::none(),
+        );
+        let new_expiration_time_sec = expiration_time_sec + time_helper::years_to_seconds(1);
+        let now = timestamp::now_seconds();
+        assert!(new_expiration_time_sec >= now, error::invalid_state(EMIGRATION_ALREADY_EXPIRED));
```
angieyth, line 786: if the original expiration date passed, should we allow them to migrate? I think we should check the original date instead of the extended date?

0xaptosj: +1 unless we want to give some grace period, otherwise i think if original expiration time > now then error

core_v2/sources/domains.move
```diff
@@ -935,6 +953,7 @@ module aptos_names_v2::domains {
         if (option::is_some(&target_addr)) {
             set_name_address_internal(option::none(), domain_name, *option::borrow(&target_addr));
         }
+        // TODO: If the name was a primary name in v1 we should make it a primary name in v2
```
angieyth, line 956: love this!

angieyth (approved):

PR #32, opened 2023-07-24
[not used] add new struct for manager, domain, subdomain

0xaptosj:
Introduce new structs first.

### DIFF
--- a/core_v2/sources/domains.move
+++ b/core_v2/sources/domains.move
@@ -68,7 +68,7 @@ module aptos_names_v2::domains {
 
     #[resource_group_member(group = aptos_framework::object::ObjectGroup)]
     struct DomainNameRecordTokenV2 has key {
-        name_domain_name: String,
+        domain_name: String,
         expiration_time_sec: u64,
         target_address: Option<address>,
 
@@ -78,7 +78,7 @@ module aptos_names_v2::domains {
 
     #[resource_group_member(group = aptos_framework::object::ObjectGroup)]
     struct SubdomainNameRecordTokenV2 has key {
-        name_subdomain_name: String,
+        subdomain_name: String,
         expiration_time_sec: u64,
         target_address: Option<address>,
         domain: Object<DomainNameRecordTokenV2>,
--- a/core_v2/sources/domains.move
+++ b/core_v2/sources/domains.move
@@ -71,7 +71,6 @@ module aptos_names_v2::domains {
         name_domain_name: String,
         expiration_time_sec: u64,
         target_address: Option<address>,
-        subdomain_collection_name: String,
 
         extend_ref: object::ExtendRef,
         transfer_ref: object::TransferRef,
@@ -83,6 +82,7 @@ module aptos_names_v2::domains {
         expiration_time_sec: u64,
         target_address: Option<address>,
         domain: Object<DomainNameRecordTokenV2>,
+        use_domain_expiration_time: bool,
 
         extend_ref: object::ExtendRef,
         transfer_ref: object::TransferRef,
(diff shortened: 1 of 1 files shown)


core_v2/sources/domains.move
```diff
@@ -60,6 +60,34 @@ module aptos_names_v2::domains {
...
     }
 
+    /// Manager object refs
+    struct Manager has key {
+        /// The extend_ref of the manager object to get its signer
+        extend_ref: object::ExtendRef,
+    }
+
+    #[resource_group_member(group = aptos_framework::object::ObjectGroup)]
+    struct DomainNameRecordTokenV2 has key {
```
0xaptosj, line 70: feel free to suggest better naming

core_v2/sources/domains.move
```diff
@@ -60,6 +60,34 @@ module aptos_names_v2::domains {
...
+        name_domain_name: String,
+        expiration_time_sec: u64,
+        target_address: Option<address>,
+        subdomain_collection_name: String,
```
angieyth, line 74: do we need this? or we can just use the domain name as the collection name?

core_v2/sources/domains.move
```diff
@@ -60,6 +60,34 @@ module aptos_names_v2::domains {
...
+
+        extend_ref: object::ExtendRef,
+        transfer_ref: object::TransferRef,
+    }
+
+    #[resource_group_member(group = aptos_framework::object::ObjectGroup)]
+    struct SubdomainNameRecordTokenV2 has key {
+        name_subdomain_name: String,
+        expiration_time_sec: u64,
```
angieyth, line 83: we need a expiration flag as `auto-renewal` or a better name here. 
when the flag is true, we will refer the expiration date of the domain as the expiration date for this subdomain

core_v2/sources/domains.move
```diff
@@ -60,6 +60,34 @@ module aptos_names_v2::domains {
...
+        target_address: Option<address>,
+        domain: Object<DomainNameRecordTokenV2>,
```
angieyth, line 85: I was thinking what should we store, the domain name or the object. it depends on which one we use more often(?)

core_v2/sources/domains.move
0xaptosj: agree, removed

core_v2/sources/domains.move
0xaptosj: i prefer we store object, as getting name from object is easier

core_v2/sources/domains.move
0xaptosj: i name it `use_domain_expiration_time`

core_v2/sources/domains.move
```diff
@@ -60,6 +60,34 @@ module aptos_names_v2::domains {
...
```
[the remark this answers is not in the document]
angieyth: I can think of 2:
1. get the full name of the subdomain -> we need the domain name
2. get the expiration date when it's auto-renew-> we need the domain object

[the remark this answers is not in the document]
0xaptosj:
```
    inline fun subdomain_token_addr_inline(
        domain_token: Object<DomainNameRecordTokenV2>,
        subdomain_name: String,
    ): address acquires DomainNameRecordTokenV2 {
        let domain_token_addr = object::object_address(&domain_token);
        let subdomain_collection_name = &borrow_global<DomainNameRecordTokenV2>(domain_token_addr).domain_name;
        token::create_token_address(&domain_token_addr, subdomain_collection_name, &subdomain_name)
    }
```
it's used to get the subdomain token address since it's the creator of the subdomain collection

PR #37, opened 2023-07-26
Add `view` directive to view functions

BriungRi:
[WIP] because we should refactor our view functions to not use `Option` for subdomain


0xaptosj:
what's the intension of adding `view` here?

0xaptosj (approved):

PR #38, opened 2023-07-26
[ans] adding renewal functions to domain/subdomain

angieyth:
## description
In this PR I changed the `struct NameRecord` to the one we agreed on. and add 
1. renewal function for domain. 
2. renewal function for subdomain (manual set the date when the flag `use_domain_expiration_sec` is false
3. setter function for subdomain to switch the flag `use_domain_expiration_sec`

## test
run `aptos move test` to make sure it passes all the tests

### DIFF
--- a/core_v2/sources/domains.move
+++ b/core_v2/sources/domains.move
@@ -53,6 +53,20 @@ module aptos_names_v2::domains {
     const EVALID_SIGNATURE_REQUIRED: u64 = 16;
     /// The domain is too short.
     const EDOMAIN_TOO_SHORT: u64 = 17;
+    /// The name is not expired in 6 months, thus not eligible for renewal
+    const EDOMAIN_NOT_AVAILABLE_TO_RENEW: u64 = 18;
+    /// The subdomain is not eligible for renewal
+    const ESUBDOMAIN_IS_AUTO_RENEW: u64 = 19;
+    /// The name is expired
+    const ENAME_EXPIRED: u64 = 20;
+    /// The subdomain not exist
+    const ESUBDOMAIN_NOT_EXIST: u64 = 21;
+
+
+    /// The enum for RegisterNameEventV1
+    const REGISTER_NAME_EVENT: u64 = 1;
+    const RENEW_NAME_EVENT: u64 = 2;
+
 
     /// Tokens require a signer to create, so this is the signer for the collection
     struct CollectionCapabilityV2 has key, drop {
@@ -122,10 +136,11 @@ module aptos_names_v2::domains {
     /// Includes the the fee paid for the registration, and the expiration time
     /// Also includes the so we can tell which version of a given domain NFT is the latest
     struct RegisterNameEventV1 has drop, store {
-        subdomain_name: Option<String>,
         domain_name: String,
+        subdomain_name: Option<String>,
         registration_fee_octas: u64,
         expiration_time_secs: u64,
+        event_type: u64,
     }
 
     /// This is only callable during publishing
@@ -173,7 +188,7 @@ module aptos_names_v2::domains {
             set_reverse_lookup_events: account::new_event_handle<SetReverseLookupEventV1>(account),
         });
     }
-    
+
     public fun get_token_signer_address(): address acquires CollectionCapabilityV2 {
         account::get_signer_capability_address(&borrow_global<CollectionCapabilityV2>(@aptos_names_v2).capability)
     }
@@ -427,10 +442,11 @@ module aptos_names_v2::domains {
         event::emit_event<RegisterNameEventV1>(
             &mut borrow_global_mut<RegisterNameEventsV1>(@aptos_names_v2).register_name_events,
             RegisterNameEventV1 {
-                subdomain_name,
                 domain_name,
+                subdomain_name,
                 registration_fee_octas: price,
                 expiration_time_secs: name_expiration_time_secs,
+                event_type: REGISTER_NAME_EVENT,
             },
         );
     }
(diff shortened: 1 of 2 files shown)


core_v2/sources/domains.move
```diff
@@ -505,14 +527,124 @@ module aptos_names_v2::domains {
...
+        let (is_valid, length) = utf8_utils::string_is_allowed(&domain_name);
+        assert!(is_valid, error::invalid_argument(EDOMAIN_HAS_INVALID_CHARACTERS));
+        assert!(length <= config::max_domain_length(), error::out_of_range(EDOMAIN_TOO_LONG));
+        assert!(length >= config::min_domain_length(), error::out_of_range(EDOMAIN_TOO_SHORT));
+
+        let price = price_model::price_for_domain_v1(length, num_years);
+        renew_domain_internal(sign, domain_name, num_years, price);
+    }
+
+    fun renew_domain_internal(
```
angieyth, line 553: I pass in price here so that we can call this func directly when we want to force renewal without paying price.

core_v2/sources/domains.move
```diff
@@ -505,14 +527,124 @@ module aptos_names_v2::domains {
...
     }
 
+    public fun renew_domain(
+        sign: &signer,
+        domain_name: String,
+        num_years: u8,
+    ) acquires CollectionCapabilityV2, NameRecord, RegisterNameEventsV1 {
+        // check the domain eligibility
+        let (is_valid, length) = utf8_utils::string_is_allowed(&domain_name);
+        assert!(is_valid, error::invalid_argument(EDOMAIN_HAS_INVALID_CHARACTERS));
```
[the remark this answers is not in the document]
angieyth: in the beginning I just needed the length, but then I was thinking if we add new naming rules, it'll be good to take renewal as a chance to remove those names not eligible anymore.

core_v2/sources/domains.move
```diff
@@ -140,10 +146,11 @@ module aptos_names_v2::domains {
     /// Includes the the fee paid for the registration, and the expiration time
     /// Also includes the so we can tell which version of a given domain NFT is the latest
     struct RegisterNameEventV1 has drop, store {
-        subdomain_name: Option<String>,
         domain_name: String,
+        subdomain_name: Option<String>,
         registration_fee_octas: u64,
         expiration_time_secs: u64,
+        event_type: u64,
```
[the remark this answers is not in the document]
angieyth: I'll ask Ying or Bowen for this

core_v2/sources/domains.move
```diff
@@ -140,10 +146,11 @@ module aptos_names_v2::domains {
     /// Includes the the fee paid for the registration, and the expiration time
     /// Also includes the so we can tell which version of a given domain NFT is the latest
     struct RegisterNameEventV1 has drop, store {
-        subdomain_name: Option<String>,
         domain_name: String,
+        subdomain_name: Option<String>,
```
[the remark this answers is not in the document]
angieyth: should we? this is the event entry for indexer to record. we probably still want it to be explicitly showed in the event? 
that being said, I asked Brian Gao and he suggested to have another event type for renewal. I'll add the event in a separate PR and leave this event as it is

core_v2/sources/domains.move
```diff
@@ -505,14 +527,124 @@ module aptos_names_v2::domains {
...
+        sign: &signer,
+        subdomain_name: String,
+        domain_name: String,
+        registration_duration_secs: u64,
+    ) acquires CollectionCapabilityV2, NameRecord{
+        validate_subdomain_to_renew(sign, subdomain_name, domain_name);
+
+        let record = get_record_mut(domain_name, option::some(subdomain_name));
+        // check the auto-renew flag
+        if (option::is_some(&record.subdomain_ext)) {
```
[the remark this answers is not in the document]
angieyth: if someone accidentally pass a domain into this function, they can still extend the expiration date. oh maybe this is a bad thing because there no cost for this function. I'll throw error then.

core_v2/sources/domains.move
```diff
@@ -505,14 +527,124 @@ module aptos_names_v2::domains {
...
+        let record = get_record_mut(domain_name, option::none());
+        // check if the domain is expired in 6 month
+        let six_monthes  = 6 * 30 * 24 * 60 * 60;
+        // renew the domain
+        if (record.expiration_time_sec - timestamp::now_seconds() > six_monthes) {
+            assert!(false, error::invalid_state(EDOMAIN_NOT_AVAILABLE_TO_RENEW));
+        };
+    }
+
+    /// this is for domain owner to update subdomain expiration time
```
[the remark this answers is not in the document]
angieyth: I don't think we should leave that for subdomain owners if our principle is to leave domain owner full control of subdomains. but I'll still rename in case :P

core_v2/sources/domains.move
```diff
@@ -53,6 +55,16 @@ module aptos_names_v2::domains {
     const EVALID_SIGNATURE_REQUIRED: u64 = 16;
     /// The domain is too short.
     const EDOMAIN_TOO_SHORT: u64 = 17;
+    /// The name is not expired in 6 months, thus not eligible for renewal
+    const EDOMAIN_NOT_AVAILABLE_TO_RENEW: u64 = 18;
+    /// The subdomain is not eligible for renewal
+    const ESUBDOMAIN_IS_AUTO_RENEW: u64 = 19;
```
0xaptosj, line 61: can we rename this to `ESUBDOMAIN_IS_NOT_AVAILABLE_AUTO_RENEW`

core_v2/sources/domains.move
```diff
@@ -53,6 +55,16 @@ module aptos_names_v2::domains {
...
+    /// The name is expired
+    const ENAME_EXPIRED: u64 = 20;
+    /// The subdomain not exist
+    const ESUBDOMAIN_NOT_EXIST: u64 = 21;
+    /// The subdomain not exist
+    const ENOT_SUBDOMAIN: u64 = 22;
```
0xaptosj, line 67: what's the difference between `ESUBDOMAIN_NOT_EXIST` and `ENOT_SUBDOMAIN`?

core_v2/sources/domains.move
```diff
@@ -517,6 +528,130 @@ module aptos_names_v2::domains {
...
+    }
+
+    fun validate_domain_name_string(
+        domain_name: String,
+    ): (bool, u64) {
+        let (is_valid, length) = utf8_utils::string_is_allowed(&domain_name);
+        assert!(is_valid, error::invalid_argument(EDOMAIN_HAS_INVALID_CHARACTERS));
+        assert!(length <= config::max_domain_length(), error::out_of_range(EDOMAIN_TOO_LONG));
+        assert!(length >= config::min_domain_length(), error::out_of_range(EDOMAIN_TOO_SHORT));
+        return (is_valid, length)
```
0xaptosj, line 575: can we only return `length`? since we never use `is_valid` outside this function

core_v2/sources/domains.move
```diff
@@ -517,6 +528,130 @@ module aptos_names_v2::domains {
...
+    }
+
+    public fun is_domain_in_renewal_window(
+        domain_name: String,
+    ): bool acquires CollectionCapabilityV2, NameRecordV2 {
+        // check if the domain is registered
+        assert!(name_is_registered(option::none(), domain_name), error::not_found(ENAME_NOT_EXIST));
+        // check if the domain is expired already
+        assert!(!name_is_expired(option::none(), domain_name), error::invalid_state(ENAME_NOT_AVAILABLE));
```
0xaptosj, line 584: should this return  something like `ENAME_EXPIRED`?

core_v2/sources/domains.move
```diff
@@ -517,6 +528,130 @@ module aptos_names_v2::domains {
...
+        let record = get_record_mut(domain_name, option::none());
+
+        record.expiration_time_sec > timestamp::now_seconds() + MAX_REMAINING_TIME_FOR_RENEWAL_SEC
+    }
+
+    /// this is for domain owner to update subdomain expiration time
+    public fun renew_subdomain_as_domain_owner(
```
0xaptosj, line 591: since `renew_subdomain_as_domain_owner` is an entry function called directly by external parties, we should do some permission check, right? like checking signer is the subdomain's domain owner

core_v2/sources/domains.move
```diff
@@ -517,6 +528,130 @@ module aptos_names_v2::domains {
...
+        // check the auto-renew flag
+        let subdomain_ext = option::borrow(&record.subdomain_ext);
+        if (subdomain_ext.use_domain_expiration_sec) {
+            assert!(false, error::invalid_state(ESUBDOMAIN_IS_AUTO_RENEW));
+        };
+        // manually set the expiration date
+        record.expiration_time_sec = record.expiration_time_sec + registration_duration_secs;
+    }
+
+    public fun set_subdomain_renewal_policy(
```
0xaptosj, line 611: also need to check signer is subdomain's domain owner i believe

core_v2/sources/domains.move
angieyth: this error is about that this subdomain is auto renew so we can't manually set a expiration date on it. I also think the name is a bit bad, but can't think of a better one. any suggestion?

0xaptosj: i seem maybe `ESUBDOMAIN_IS_AUTO_RENEW` is fine

core_v2/sources/domains.move
angieyth: The name is not a subdomain. sorry forgot to change the comment :P

core_v2/sources/domains.move
```diff
@@ -517,6 +528,131 @@ module aptos_names_v2::domains {
...
+
+    fun renew_domain_internal(
+        sign: &signer,
+        domain_name: String,
+        num_years: u8,
+        price: u64,
+    ) acquires CollectionCapabilityV2, NameRecordV2, RegisterNameEventsV1 {
+        let record = get_record_mut(domain_name, option::none());
+        let registration_duration_secs = time_helper::years_to_seconds((num_years as u64));
+        record.expiration_time_sec = timestamp::now_seconds() + registration_duration_secs;
```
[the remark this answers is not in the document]
angieyth: you are right!!! I thought I changed this already

PR #43, opened 2023-08-01
Change behavior of name_is_expired to not error out

BriungRi:
This change creates an implication that all records already exist, but by default they are expired.

It also simplifies the behavior when ANS v2 needs to look into ANS v1 to know whether a name is expired. Since we can't catch errors in Move, it is simpler to not unnecessarily error for a view function like name_is_expired

# Test Plan

Coverage of all expiration behavior

1. New test for `name_is_expired` for an unregistered name + subdomain
2. Existing test coverage in `test_helper::register_name` for `name_is_expired` right after registration
3. Existing test coverage in `names_are_registerable_after_expiry_e2e_test` that a registered name eventually expires


0xaptosj:
tag @CapCap to see if this have any impact

movekevin (approved):

PR #46, opened 2023-08-02
[ans] set subdomain expiration 

angieyth:
in this PR
1. use `set_subdomain_expiration` instead of renew to give domain owner more flexibility
2. remove num_years and use registration_duration_secs instead for the params

### DIFF
--- a/core_v2/sources/domains.move
+++ b/core_v2/sources/domains.move
@@ -73,7 +73,9 @@ module aptos_names_v2::domains {
     const ENOT_A_SUBDOMAIN: u64 = 22;
     /// The domain expiration, even after migration extension, is past now.
     const EMIGRATION_ALREADY_EXPIRED: u64 = 23;
+    /// The subdomain expiration can be set any date before the domain expiration
     const ESUBDOMAIN_EXPIRATION_PASS_DOMAIN_EXPIRATION: u64 = 24;
+    /// The duration must be whole years
     const EDURATION_MUST_BE_WHOLE_YEARS: u64 = 25;
 
     /// Tokens require a signer to create, so this is the signer for the collection
--- a/core_v2/sources/domains.move
+++ b/core_v2/sources/domains.move
@@ -29,6 +29,7 @@ module aptos_names_v2::domains {
     const MAX_REMAINING_TIME_FOR_RENEWAL_SEC: u64 = 15552000;
     /// 2024/03/07 23:59:59
     const AUTO_RENEWAL_EXPIRATION_CUTOFF_SEC: u64 = 1709855999;
+    const SECONDS_PER_YEAR: u64 = 60 * 60 * 24 * 365;
 
     /// The Naming Service contract is not enabled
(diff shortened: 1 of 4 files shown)


core_v2/sources/domains.move
```diff
@@ -344,9 +344,10 @@ module aptos_names_v2::domains {
     fun register_domain_generic(
         sign: &signer,
         domain_name: String,
-        num_years: u8
+        registration_duration_secs: u64,
     ) acquires CollectionCapabilityV2, NameRecordV2, RegisterNameEventsV1, ReverseRecord, SetNameAddressEventsV1, SetReverseLookupEventsV1 {
         assert!(config::is_enabled(), error::unavailable(ENOT_ENABLED));
+        let num_years = (time_helper::seconds_to_years(registration_duration_secs) as u8);
         assert!(
             num_years > 0 && num_years <= config::max_number_of_years_registered(),
```
0xaptosj, line 352: should we also change `max_number_of_years_registered` to second?

0xaptosj: ok price is still based on year, do we want to change that too?

angieyth: I was thinking but not sure which one is better

core_v2/sources/domains.move
```diff
@@ -578,24 +576,24 @@ module aptos_names_v2::domains {
     public entry fun renew_domain(
```
0xaptosj, line 576: how about rename function name to `set_domain_expiration` or `update_domain_expiration`?

core_v2/sources/domains.move
angieyth: I wouldn't want to change the price model while migration, that's why I didn't touch that part. Although we change the interface here, but we probably want to keep the current price for a while?

core_v2/sources/domains.move
angieyth: we still `renew` domains within the renewable window, but we `set` subdomain expirations to give domain owner more flexibility.

0xaptosj (approved):

core_v2/sources/domains.move
```diff
@@ -636,11 +634,11 @@ module aptos_names_v2::domains {
...
     /// this is for domain owner to update subdomain expiration time
-    public fun renew_subdomain_as_domain_owner(
+    public fun set_subdomain_expiration_as_domain_owner(
         sign: &signer,
         subdomain_name: String,
         domain_name: String,
-        registration_duration_secs: u64,
+        expiration_time_sec: u64,
     ) acquires CollectionCapabilityV2, NameRecordV2{
         validate_subdomain_to_renew(sign, subdomain_name, domain_name);
```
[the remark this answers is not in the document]
angieyth: what's the better name here? now the time stamp is a absolute timestamp instead of a time span. this name was suggested by copilot :P I couldn't think of a better one.

angieyth:
the subdomain policy is still something to be changed. I'll revise that part when we have the policy done!

PR #47, opened 2023-08-04
add view functions

0xaptosj:
Add 3 view functions `get_domain_props`, `get_subdomain_props`, `get_reverse_lookup_name`

### DIFF
--- a/core_v2/sources/domains.move
+++ b/core_v2/sources/domains.move
@@ -1124,6 +1124,39 @@ module aptos_names_v2::domains {
         (extract_subdomain_name(record), record.domain_name)
     }
 
+    #[view]
+    public fun get_name_record_v1_props_for_domain(
+        domain: String,
+    ): (u64, Option<address>) acquires CollectionCapabilityV2, NameRecordV2 {
+        get_name_record_v1_props_for_name(option::none(), domain)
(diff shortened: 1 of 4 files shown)


core_v2/sources/domains.move
```diff
@@ -1108,6 +1108,39 @@ module aptos_names_v2::domains {
...
+        subdomain: String,
+        domain: String,
+    ): (u64, Option<address>) acquires CollectionCapabilityV2, NameRecordV2 {
+        get_name_record_v1_props_for_name(option::some(subdomain), domain)
+    }
+
+    #[view]
+    public fun get_reverse_lookup_name(
+        account_addr: address
+    ): Option<String> acquires ReverseRecord, NameRecordV2 {
```
0xaptosj, line 1129: this will return `hello.apt` if lookup result is a domain, instead of just `hello`

0xaptosj:
>  If we add more custom logic for expirations (subdomain policies), we can just update get_subdomain_expiration as opposed to having to overload get_name_record_v1_props_for_name

@BriungRi are you saying we should deprecate `get_name_record_v1_props_for_name` and use `get_domain_expiration` and `get_domain_target` instead?

PR #48, opened 2023-08-04
Setup ANS v1 for all ANS v2 tests

BriungRi:


core_v2/sources/domain_e2e_tests.move
```diff
@@ -597,13 +917,7 @@ module aptos_names_v2::domain_e2e_tests {
         rando: signer,
         foundation: signer
     ) {
-        // Setup test env for v1
-        let users = aptos_names::test_helper::e2e_test_setup(aptos_names, user, &aptos, rando, &foundation);
-
-        // Setup test env for v2
-        account::create_account_for_test(@aptos_names_v2);
-        aptos_names_v2::domains::init_module_for_test(aptos_names_v2);
-        aptos_names_v2::config::set_fund_destination_address_test_only(signer::address_of(&foundation));
```
0xaptosj, line 606: since we are calling `set_fund_destination_address_test_only()` in `e2e_test_setup()` below, does that mean we are actually duplicating the operation here?

0xaptosj (approved):

core_v2/sources/test_helper.move
```diff
@@ -46,11 +46,21 @@ module aptos_names_v2::test_helper {
...
+        aptos: &signer,
+        rando: signer,
+        foundation: &signer
+    ): vector<signer> {
+        account::create_account_for_test(@aptos_names);
         account::create_account_for_test(@aptos_names_v2);
         let new_accounts = setup_and_fund_accounts(aptos, foundation, vector[user, rando]);
         timestamp::set_time_has_started_for_testing(aptos);
-        domains::init_module_for_test(myself);
+        aptos_names::domains::init_module_for_test(aptos_names);
```
0xaptosj, line 62: is v1 used anywhere? only place i found is setting it up here

core_v2/sources/domain_e2e_tests.move
0xaptosj: i see thanks for explanation

PR #49, opened 2023-08-04
[ans] adding transfer subdomain function

angieyth:
## Description

a domain can transfer a subdomain whether they are a subdomain owner or not. 
a target address is also passed in when transferring ownership

## Test plan
`aptos move test --filter test_transfer_subdomain`

### DIFF
--- a/core_v2/sources/domains.move
+++ b/core_v2/sources/domains.move
@@ -639,13 +639,13 @@ module aptos_names_v2::domains {
             error::out_of_range(EINVALID_NUMBER_YEARS)
         );
     }
-    public entry fun transfer_subdomain_owner(
+    public fun transfer_subdomain_owner(
         sign: &signer,
         subdomain_name: String,
         domain_name: String,
         new_owner_address: address,
-        new_target_address: address,
-    ) acquires CollectionCapabilityV2, NameRecordV2 {
+        new_target_address: Option<address>,
+    ) acquires CollectionCapabilityV2, NameRecordV2, ReverseRecord, SetReverseLookupEventsV1 {
         // validate user own the domain
         let signer_addr = signer::address_of(sign);
         assert!(
@@ -655,8 +655,10 @@ module aptos_names_v2::domains {
 
         let token_addr = token_addr_inline(domain_name, option::some(subdomain_name));
         let record = borrow_global_mut<NameRecordV2>(token_addr);
-        record.target_address = option::some(new_target_address);
+        record.target_address = new_target_address;
         object::transfer_with_ref(object::generate_linear_transfer_ref(&record.transfer_ref), new_owner_address);
+        // clear the primary name
+        clear_reverse_lookup_for_name(option::some(subdomain_name), domain_name);
     }
 
     fun validate_name_string(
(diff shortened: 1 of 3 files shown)


core_v2/sources/domains.move
```diff
@@ -628,14 +628,36 @@ module aptos_names_v2::domains {
...
             num_years > 0 && num_years <= config::max_number_of_years_registered(),
             error::out_of_range(EINVALID_NUMBER_YEARS)
         );
     }
+    public entry fun transfer_subdomain_owner(
+        sign: &signer,
+        subdomain_name: String,
+        domain_name: String,
+        new_owner_address: address,
+        new_target_address: address,
```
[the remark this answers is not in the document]
angieyth: I intentionally made it non-optional so that it can be an entry function. what do you think?

0xaptosj, line 647: what if i want to transfer subdomain to myself (the domain owner) and remove the target address? basically reclaim the subdomain and make it available

core_v2/sources/domains.move
```diff
@@ -628,14 +628,36 @@ module aptos_names_v2::domains {
...
+            error::invalid_argument(EDURATION_MUST_BE_WHOLE_YEARS)
+        );
 
         let num_years = (time_helper::seconds_to_years(registration_duration_secs) as u8);
         assert!(
             num_years > 0 && num_years <= config::max_number_of_years_registered(),
             error::out_of_range(EINVALID_NUMBER_YEARS)
         );
     }
+    public entry fun transfer_subdomain_owner(
```
0xaptosj, line 642: do we need to set reverse lookup in any case?

angieyth: oh you mean if it was someone's primary name we should remove it?

core_v2/sources/domains.move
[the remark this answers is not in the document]
angieyth: I'll add those two to the task list and come back when Jianyi's work done.

core_v2/sources/domains.move
```diff
@@ -628,14 +628,38 @@ module aptos_names_v2::domains {
...
+            error::invalid_argument(EDURATION_MUST_BE_WHOLE_YEARS)
+        );
 
         let num_years = (time_helper::seconds_to_years(registration_duration_secs) as u8);
         assert!(
             num_years > 0 && num_years <= config::max_number_of_years_registered(),
             error::out_of_range(EINVALID_NUMBER_YEARS)
         );
     }
+    public fun transfer_subdomain_owner(
```
0xaptosj, line 642: nit: rename it to `transfer_subdomain_as_domain_owner`? we have a `set_subdomain_expiration_as_domain_owner`

core_v2/sources/subdomain_e2e_tests.move
```diff
@@ -182,6 +182,105 @@ module aptos_names_v2::subdomain_e2e_tests {
...
+
+        // user is the owner of domain
+        let is_owner = domains::is_owner_of_name(user_addr, option::none(), test_helper::domain_name());
+        assert!(is_owner, 1);
+
+        // transfer the subdomain to rando
+        domains::transfer_subdomain_owner(user, test_helper::subdomain_name(), test_helper::domain_name(), rando_addr, option::some(rando_addr));
+
+        {
+            // when rando owns the subdomain but not the domain, rando can't transfer subdomain ownership.
```
0xaptosj, line 273: although rando can't call `transfer_subdomain_owner` to transfer it, it can still use the built in transfer, right? unless we disable the built in transfer

PR #51, opened 2023-08-04
Add lookup from v2 into v1 for `name_is_registerable`

BriungRi:


angieyth:
is this rebased properly? the test changes was in https://github.com/aptos-labs/aptos-names-contracts/pull/48/files?

PR #58, opened 2023-08-08
[ans] adding force_set_name_expiration function

angieyth:
## Description
adding the force_renew_domain function
move the paying price part out of renew_domain_internal

## Test
`aptos move test`


0xaptosj (approved):
add entry to `force_set_name_expiration`

PR #59, opened 2023-08-09
Allow setting target address and transfer to address during registration

0xaptosj:
**Summary**
1. rename `set_reverse_lookup` to `set_name_address_and_reverse_lookup` to avoid confusion.
2. change the logic on when to `set_name_address_and_reverse_lookup`

|previous behavior | new behavior|
|-|-|
|if signer doesn't have a primary name (either due to first time buying a name, or had once but cleared) | on top of previous behavior, we check another condition `should_set_reverse_lookup_and_target_addres_result`, this returns true only when signer set both target address and transfer to address to itself, then we set primary name |

```
    public fun should_set_reverse_lookup_and_target_address(
        signer_address: address,
        target_address: Option<address>,
        transfer_to_address: Option<address>,
    ): bool {
        let transfer_to_address = option::get_with_default(&transfer_to_address, signer_address);
        let target_address = option::get_with_default(&target_address, signer_address);
        if (target_address == signer_address && transfer_to_address == signer_address) {
            true
        } else {
            false
        }
    }
```

3. change the logic on when to set target address

|previous behavior | new behavior|
|-|-|
|if primary name is set, target address will be set along, or user is registering a domain. in both cases, target address is set to the signer | on top of previous behavior, also set target address if it's passed explicitly, in this case even signer is registering a subdomain we still set target address |

```
    public fun should_set_target_address_in_register_name(
        subdomain_name: Option<String>,
        target_address: Option<address>,
    ): bool {
        if (!is_subdomain(subdomain_name)) {
            // If signer is registering a domain, automatically set the name to point to target address, if not set use signer address
            true
        }
        else if (option::is_some(&target_address)) {
            // Else if target address is explicitly provided, set it
            true
        }
        else {
            // Else signer is registering a subdomain and target address not explicitly set, leave it as none
            false
        }
    }
```


angieyth:
do you think we should add an unit test for transfer and another one for target address? (I sound like brian :P)

0xaptosj:
> do you think we should add an unit test for transfer and another one for target address? (I sound like brian :P)

I should, im figuring out how, your are right unit test is the most painful thing

core_v2/sources/domains.move
```diff
@@ -1047,6 +1103,8 @@ module aptos_names_v2::domains {
             domain_name,
             new_expiration_time_sec - now,
             0,
+            option::none(),
+            option::none(),
         );
         // TODO: `register_name_internal` should accept a `target_addr`
```
angieyth, line 1109: maybe remove this?

core_v2/sources/domains.move
```diff
@@ -1173,6 +1231,21 @@ module aptos_names_v2::domains {
         timestamp::now_seconds() >= expiration_time_sec
     }
 
+    /// Given signer, target address and transfer to address, return true if we should set name address and reverse lookup, false otherwise
+    public fun should_set_reverse_lookup_and_target_address(
```
angieyth, line 1235: should it be target address or name address? what's the difference between setting name address and reverse lookup?

core_v2/sources/test_helper.move
```diff
@@ -85,14 +95,40 @@ module aptos_names_v2::test_helper {
         let set_name_address_event_v1_event_count_before = domains::get_set_name_address_event_v1_count();
         let set_reverse_lookup_event_v1_event_count_before = domains::get_set_reverse_lookup_event_v1_count();
 
+        let should_set_reverse_lookup_and_target_address_result_result = domains::should_set_reverse_lookup_and_target_address(
```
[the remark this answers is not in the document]
0xaptosj: fixed

core_v2/sources/domains.move
0xaptosj: we need to set them to satisfy the function signature

core_v2/sources/domains.move
0xaptosj: name_address is the same thing as target_address, they are used interchangeablely in the code

core_v2/sources/domains.move
angieyth: I mean the comment

core_v2/sources/domains.move
```diff
@@ -471,31 +499,53 @@ module aptos_names_v2::domains {
...
+        let is_target_address_unset = true;
         let reverse_lookup_result = get_reverse_lookup(account_addr);
         if (option::is_none(&reverse_lookup_result)) {
-            // If the user has no reverse lookup set, set the user's reverse lookup.
-            set_reverse_lookup(sign, subdomain_name, domain_name);
-        } else if (!is_subdomain(subdomain_name)) {
-            // Automatically set the name to point to the sender's address
-            set_target_address_internal(subdomain_name, domain_name, signer::address_of(sign));
+            // If the signer has no reverse lookup set and signer is minting for itself, set the user's reverse lookup and target address.
+            if (result_of_should_set_reverse_lookup_and_target_address) {
```
angieyth, line 538: is it possible to separate the logic of reverse look up and set target address in `should_set_reverse_lookup_and_target_address_in_register_name`

I see we also have `should_set_target_address_in_register_name` and it's confusing

0xaptosj: it's because set reverse lookup will always set target address at the same time, if i break it into 2, i need to break [`set_target_address_and_reverse_lookup`](https://github.com/aptos-labs/aptos-names-contracts/pull/59/files#diff-e92ae37b241c364f4ff602d97ffe2ac754cacda7ca83e29f4e44a4d154977ea9R1041)

angieyth: I do think we should break `set_target_address_and_reverse_lookup` :P 
because those are basically two different things. aren't them? 
you set primary name when the buyer is the owner
you set target address whenever there's a value passing in?

core_v2/sources/domains.move
```diff
@@ -1171,6 +1232,40 @@ module aptos_names_v2::domains {
...
+        }
+    }
+
+    /// Given signer, target address and transfer to address, return true if we should set target address, false otherwise
+    public fun should_set_target_address_in_register_name(
+        subdomain_name: Option<String>,
+        target_address: Option<address>,
+    ): bool {
+        if (!is_subdomain(subdomain_name)) {
+            // If signer is registering a domain, automatically set the name to point to target address, if not set use signer address
```
angieyth, line 1256: I don't understand this case, can you elaborate?

0xaptosj: updated the comment, basically when registering a domain, we always set the target address of the new domain

PR #61, opened 2023-08-09
[ans] default expiration policy as manual renew

angieyth:
set the default subdomain expiration policy as manual renewal (match the former behavior)


0xaptosj (approved):

0xaptosj:
maybe we should copy over the policy enum to make it more readable

angieyth:
I thought about copy over, but I feel like that'll create two sources of truth, if I change the constant in the code but don't change here, the unit tests will become useless. Thus I'll have to sacrifice readability here

PR #65, opened 2023-08-09
Separate set reverse lookup and target address to 2 functions call in register name

0xaptosj:


angieyth:
If I want to set a primary name, the target address has to be pointing to myself. by this change, can we avoid the case that I get a name for the first time but set target address to others. -> not set primary name?

core_v2/sources/domains.move
```diff
@@ -491,11 +490,12 @@ module aptos_names_v2::domains {
 
         let reverse_lookup_result = get_reverse_lookup(account_addr);
         if (option::is_none(&reverse_lookup_result)) {
-            // If the user has no reverse lookup set, set the user's reverse lookup.
-            set_reverse_lookup(sign, subdomain_name, domain_name);
+            // If the user has no reverse lookup set, set the user's reverse lookup and target address.
+            set_target_address(sign, subdomain_name, domain_name, account_addr);
```
angieyth, line 494: this can combine with line 498?

angieyth (approved):

core_v2/sources/domains.move
```diff
@@ -489,13 +488,15 @@ module aptos_names_v2::domains {
             );
         };
 
-        let reverse_lookup_result = get_reverse_lookup(account_addr);
-        if (option::is_none(&reverse_lookup_result)) {
-            // If the user has no reverse lookup set, set the user's reverse lookup.
-            set_reverse_lookup(sign, subdomain_name, domain_name);
-        } else if (!is_subdomain(subdomain_name)) {
-            // Automatically set the name to point to the sender's address
-            set_target_address_internal(subdomain_name, domain_name, signer::address_of(sign));
```
0xaptosj, line 498: `set_target_address_internal` could be replace by [`set_target_address`](https://github.com/aptos-labs/aptos-names-contracts/blob/bl/rew/core_v2/sources/domains.move#L873)  because `set_target_address` does 2 more things beside `set_target_address_internal`, 1. it will clear the reverse lookup if it's a domain (but it's not applicable here, cause the new name doesn't have a target address yet,  2. clear signer's reverse lookup if signer's reverse lookup is the name registering now, this is also not applicable cause we always set target address first before set reverse lookup

0xaptosj: however this is pretty dangerous, let me know if you guys think i should make this change `set_target_address_internal` -> `set_target_address`

core_v2/sources/domains.move
```diff
@@ -489,13 +488,15 @@ module aptos_names_v2::domains {
             );
         };
 
-        let reverse_lookup_result = get_reverse_lookup(account_addr);
-        if (option::is_none(&reverse_lookup_result)) {
-            // If the user has no reverse lookup set, set the user's reverse lookup.
-            set_reverse_lookup(sign, subdomain_name, domain_name);
```
0xaptosj, line 495: `set_reverse_lookup` internally calls `set_target_address` and `set_reverse_lookup_internal`

0xaptosj:
Close this one cause we will support register name, set target address and transfer using a new entry function

PR #66, opened 2023-08-10
Remove v1 dependency from core_v2

BriungRi:
With router, v2 no longer needs to know anything about v1. We rely on the router to do all the v1 checking


0xaptosj:
does that mean the migration functions will live in router?

0xaptosj (approved):

PR #67, opened 2023-08-10
Add registration, migration, and other functions for router

BriungRi:
# Description

core_v2 changes:

- Authenticate all core_v2 registrations behind a `router_signer`. This ensures that only the router is allowed to call registration. We do this because we now make the router responsible for checking that the name does not already exist in v1
  - This comprises all the changes in `core_v2/sources/domain_e2e_tests.move`, `core_v2/sources/subdomain_e2e_tests.move` and `core_v2/sources/test_helper.move`
- Move `primary_name` and `target_addr` assignment outside of `register_name_internal` and directly into `register_domain_generic` and `register_subdomain`
- Fix `validate_name_string` in `register_subdomain`. Leave TODO for unit tests
- Add `register_name_with_router` that allows the router to register the name for free. It does no automatic assignment of `primary_name` and `target_addr` because it handles the `primary_name` and `target_addr` migration in the router

router changes:

- Use resource account instead of object. Object is actually possible with `aptos_account::create_account` (required in order to store token v1 tokens), but resource account is just simpler for now.
- For registration, migration, transfers, several set_*, several get_*, implement for `MODE_V1` and `MODE_V1_AND_V2`
- Add #[view] tags

### DIFF
--- a/core_v2/sources/domains.move
+++ b/core_v2/sources/domains.move
@@ -27,7 +27,6 @@ module aptos_names_v2::domains {
     /// current MAX_REMAINING_TIME_FOR_RENEWAL_SEC is 6 months
     const MAX_REMAINING_TIME_FOR_RENEWAL_SEC: u64 = 15552000;
     /// 2024/03/07 23:59:59
-    const AUTO_RENEWAL_EXPIRATION_CUTOFF_SEC: u64 = 1709855999;
     const SECONDS_PER_YEAR: u64 = 60 * 60 * 24 * 365;
 
     /// enums for subdomain expiration policy. update validate_subdomain_expiration_policy() when adding more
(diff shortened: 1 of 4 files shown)


core_v2/sources/domains.move
```diff
@@ -377,25 +379,39 @@ module aptos_names_v2::domains {
...
         sign: &signer,
         domain_name: String,
         registration_duration_secs: u64,
     ) acquires CollectionCapabilityV2, NameRecordV2, RegisterNameEventsV1, ReverseRecord, SetNameAddressEventsV1, SetReverseLookupEventsV1 {
+        assert!(address_of(router_signer) == @router_signer, error::permission_denied(ENOT_ROUTER));
         assert!(config::unrestricted_mint_enabled(), error::permission_denied(EVALID_SIGNATURE_REQUIRED));
         register_domain_generic(sign, domain_name, registration_duration_secs);
     }
 
     public entry fun register_domain_with_signature(
```
angieyth, line 407: should those still be `public entry fun`? should we have any public entry fun in v1 and v2?

core_v2/sources/domains.move
```diff
@@ -418,8 +436,8 @@ module aptos_names_v2::domains {
         );
 
         // We are registering a subdomain name: this has no cost, but is only doable by the owner of the domain
-
-        validate_name_string(domain_name);
+        // TODO: Unit tests for trying to register invalid domains/subdomains
```
angieyth, line 439: thank you!

router/sources/router.move
```diff
@@ -0,0 +1,625 @@
...
+
+            // Calculate new expiration
+            let now = timestamp::now_seconds();
+            assert!(expiration_time_sec >= now, error::invalid_state(EMIGRATION_ALREADY_EXPIRED));
+            let new_expiration_time_sec = if (option::is_none(
+                &subdomain_name
+            ) && expiration_time_sec <= AUTO_RENEWAL_EXPIRATION_CUTOFF_SEC) {
+                expiration_time_sec + SECONDS_PER_YEAR
+            } else {
+                expiration_time_sec
```
0xaptosj, line 281: we don't renew subdomain?

PR #70, opened 2023-08-10
[ans] adding the invalid subdomain string unit test

angieyth:


core_v2/sources/domains.move
```diff
@@ -425,7 +425,7 @@ module aptos_names_v2::domains {
 
         // We are registering a subdomain name: this has no cost, but is only doable by the owner of the domain
 
-        validate_name_string(domain_name);
+        validate_name_string(subdomain_name);
```
0xaptosj, line 428: do we meed to validate both domain and subdomain?

angieyth: at this point we should have this domain created? shouldn't need to validate the domain(?)

0xaptosj: you are right

0xaptosj (approved):

PR #79, opened 2023-08-11
Implement domain renewal in router

BriungRi:


core_v2/sources/domains.move
```diff
@@ -636,26 +636,26 @@ module aptos_names_v2::domains {
     public entry fun renew_domain(
```
angieyth, line 636: maybe not a `public entry fun` anymore?

angieyth (approved):

PR #83, opened 2023-08-11
Implement get_subdomain_expiration_policy

BriungRi:


router/sources/router.move
```diff
@@ -527,19 +527,23 @@ module router::router {
...
+        domain_admin: &signer,
+        domain_name: String,
+        subdomain_name: String,
+        expiration_policy: u8,
     ) acquires RouterConfig {
         let mode = get_mode();
         if (mode == MODE_V1) {
             // Will not be implemented in v1
             abort error::not_implemented(ENOT_IMPLEMENTED_IN_MODE)
         } else if (mode == MODE_V1_AND_V2) {
```
angieyth, line 540: do you think we should have a `mode == MODE_V2` in place now so that we don't have to revise when V1 retires

angieyth (approved):

PR #88, opened 2023-08-16
Always set target address for subdomain

0xaptosj:
During subdomain registration, we should always set target address to sender, just like domain.


core_v2/sources/test_helper.move
```diff
@@ -82,14 +82,30 @@ module aptos_names_v2::test_helper {
...
+            let result = domains::get_reverse_lookup(*option::borrow(&maybe_target_address));
+            // When domain and subdomain both have target address set to user, and domain is primary name
+            // get_reverse_lookup returns the address of domain token
+            // we need to make sure the name token address match the name token address returnes from reverse lookup
+            // to know the name does have reverse look up (i.e. name is primary name)
+            if (result == option::some(domains::token_addr(domain_name, subdomain_name))) {
+                result
+            } else {
+                option::none()
+            }
```
0xaptosj, line 99: this is the most important change in this pr, after i set target address for subdomain the unit test keeps failing because it thinks `name_reverse_lookup_before` exist for subdomain now, actually it doesn't , `name_reverse_lookup_before` only be true when that name is a primary name, not when user has a primary name, that is what `user_reverse_lookup_before` for.

PR #89, opened 2023-08-16
Add reregistration_grace_sec to core_v1

BriungRi:
We want to allow a grace period after a name has expired before anyone can re-register it. This diff adds a new config key for this setting and adds two tests: (1) Try registering while during the grace period (2) Try registering after the grace period


0xaptosj (approved):
1 nit

PR #91, opened 2023-08-21
Update reverse lookup events to emit prev_*

BriungRi:
Motivation for this PR is to simplify indexing on primary name changes

1. Left TODO because noticed that we might not be accounting for dynamic subdomain expirations
2. Updated RenewNameEvent to emit `target_addr` and `is_primary_name`
3. Updated SetReverseLookupEvent to emit differently. Specifically, `account_addr` should always be the same for the same user, and we now track the delta from prev => curr state.

### DIFF
--- a/core_v2/sources/domains.move
+++ b/core_v2/sources/domains.move
@@ -129,6 +129,7 @@ module aptos_names_v2::domains {
         /// The address this reverse lookup belongs to
         account_addr: address,
 
+        /// Indexer needs knowledge of previous state
         prev_domain_name: Option<String>,
         prev_subdomain_name: Option<String>,
         prev_expiration_time_secs: Option<u64>,


core_v2/sources/domains.move
```diff
@@ -131,9 +131,14 @@ module aptos_names_v2::domains {
     /// the reverse lookup has been cleared (in which case |target_address|
     /// will be none)
     struct SetReverseLookupEvent has drop, store {
```
rtso, line 133: nit: could we rename `next_` to `curr_` to represent that this is the current state after the txn is submitted. 

Also could we include prev_ and curr_ expiration_time_secs this event.

core_v2/sources/domains.move
```diff
@@ -131,9 +131,14 @@ module aptos_names_v2::domains {
...
+        prev_subdomain_name: Option<String>,
+        prev_target_address: Option<address>,
+
+        next_domain_name: Option<String>,
+        next_subdomain_name: Option<String>,
+        next_target_address: Option<address>,
+
     }
 
     /// A name (potentially subdomain) has had it's address changed
```
rtso, line 144: For `RenewNameEvent`, can we also add something like `is_primary`

core_v2/sources/domains.move
```diff
@@ -1161,14 +1217,24 @@ module aptos_names_v2::domains {
     }
 
     fun emit_set_reverse_lookup_event_v1(
```
angieyth, line 1219: do you want to remove the v1 here too?

angieyth (approved):

rtso (approved):
Nice, lgtm! :)

PR #92, opened 2023-08-22
clean up

angieyth:
In this PR I 
1) unify the function naming as verb first like `name_is_registered` to `is_name_registered`
2) adding `clear_target_address` to router
3) remove all the `public entry fun` in `domains.move` 
4) except all the force function. 
3) move functions around in domains.move of core_v2 to make it more readable. 

my overall strategy is categorizing functions as 
  - register names 
  - primary names
  - events
  -  force functions
  - helper functions


core_v2/sources/domains.move
```diff
@@ -210,24 +207,12 @@ module aptos_names_v2::domains {
             option::none(),
             utf8(COLLECTION_URI),
         );
-        // TODO: figure out description for subdomain collection
```
angieyth, line 213: no need

core_v2/sources/domains.move
```diff
@@ -80,12 +81,6 @@ module aptos_names_v2::domains {
         burn_signer_capability: SignerCapability,
     }
 
-    /// Manager object refs
```
angieyth, line 83: no need

0xaptosj:
is it possible to break this into smaller pr :)

PR #114, opened 2023-08-30
[ans] remove the set primary name after register name

angieyth:
1. remove the logic to set primary name in domains.move
2. remove the unit tests for domain.move
3. adding unit test for router


core_v2/sources/domain_e2e_tests.move
```diff
@@ -456,53 +456,6 @@ module aptos_names_v2::domain_e2e_tests {
...
 
-    #[test(
-        router_signer = @router_signer,
-        aptos_names_v2 = @aptos_names_v2,
-        user = @0x077,
-        aptos = @0x1,
-        rando = @0x266f,
-        foundation = @0xf01d
-    )]
-    fun test_admin_force_seize_domain_name_doesnt_clear_unrelated_primary_name(
```
0xaptosj, line 467: is it possible to keep this test and fix it?

router/sources/primary_name_tests.move
```diff
@@ -44,6 +44,37 @@ module router::primary_name_tests {
...
+        router::init_module_for_test(router);
+        let users = test_helper::e2e_test_setup(aptos_names, aptos_names_v2, user1, &aptos, user2, &foundation);
+        let user = vector::borrow(&users, 0);
+        let user_addr = address_of(user);
+        let domain_name = utf8(b"test");
+        let subdomain_name = utf8(b"test");
+
+        router::register_domain(user, domain_name, SECONDS_PER_YEAR, option::none(), option::none());
+
+        let (_, primary_domain_name) = router::get_primary_name(user_addr);
```
0xaptosj, line 74: nit: also check primary_subdomain_name is none

core_v2/sources/domain_e2e_tests.move
angieyth: I was thinking about it, but I think this test is a bit weird and since we remove the primary name logic in the domains.move, we shouldn't have the test here. I added a test to the router part.

router/sources/primary_name_tests.move
```diff
@@ -44,6 +44,37 @@ module router::primary_name_tests {
...
+        let users = test_helper::e2e_test_setup(aptos_names, aptos_names_v2, user1, &aptos, user2, &foundation);
+        let user = vector::borrow(&users, 0);
+        let user_addr = address_of(user);
+        let domain_name = utf8(b"test");
+
+        router::register_domain(user, domain_name, SECONDS_PER_YEAR, option::none(), option::none());
+
+        let (primary_subdomain_name, primary_domain_name) = router::get_primary_name(user_addr);
+        assert!(*option::borrow(&primary_domain_name) == domain_name, 1);
+        assert!(option::is_none(&primary_subdomain_name), 2);
```
0xaptosj, line 75: can we also check it works after bumping mode to v1v2?

angieyth: added! the `set_primary_name_when_register()` is shared crossed modes. so I just repeat the test after I bumped the mode.

0xaptosj (approved):
ill wait for brian's sign off

PR #117, opened 2023-08-31
V1 updates for ANS v2

BriungRi:
Extract the change made to V1 from https://github.com/aptos-labs/aptos-names-contracts/pull/23 to make the PR smaller.

Changelog:
1. add new endpoint`set_reregistration_grace_sec` to allow admin set the reregistration grace period.
  a. new logic for `name_is_expired`, `name_is_registerable` based on this new grace period. Also, don't error out for `name_is_expired` if the name has never been registered before
2. add new endpoint `force_set_expiration` for admin to update the expiration of a name
3. clear primary name during `force_clear_registration`
4. let `router_signer` participate as an admin
5. change visibility of `clear_name_address` to public for easier use in the router

### DIFF
--- a/core/Move.toml
+++ b/core/Move.toml
@@ -10,10 +10,10 @@ router_signer = "_"
 
 [dependencies.AptosFramework]
 git = 'https://github.com/aptos-labs/aptos-core.git'
-rev = 'main'
+rev = 'framework-mainnet'
 subdir = 'aptos-move/framework/aptos-framework'
 
 [dependencies.AptosToken]
 git = 'https://github.com/aptos-labs/aptos-core.git'
-rev = 'main'
+rev = 'framework-mainnet'
 subdir = 'aptos-move/framework/aptos-token'
--- a/core/sources/domain_e2e_tests.move
+++ b/core/sources/domain_e2e_tests.move
@@ -604,6 +604,7 @@ module aptos_names::domain_e2e_tests {
             vector::empty<u8>()
         );
     }
+
     #[test(myself = @aptos_names, user = @0x077, aptos = @0x1, rando = @0x266f, foundation = @0xf01d)]
     fun clear_registration_property_version_test(myself: &signer, user: signer, aptos: signer, rando: signer, foundation: signer) {
         let users = test_helper::e2e_test_setup(myself, user, &aptos, rando, &foundation);


core/Move.toml
```diff
@@ -6,13 +6,14 @@ version = '1.0.0'
 aptos_names = "_"
 aptos_names_admin = "_"
 aptos_names_funds = "_"
+router_signer = "_"
 
 [dependencies.AptosFramework]
 git = 'https://github.com/aptos-labs/aptos-core.git'
-rev = 'framework-mainnet'
+rev = 'main'
```
davidiw, line 13: revert

core/Move.toml
```diff
@@ -6,13 +6,14 @@ version = '1.0.0'
...
 subdir = 'aptos-move/framework/aptos-framework'
 
 [dependencies.AptosToken]
 git = 'https://github.com/aptos-labs/aptos-core.git'
-rev = 'framework-mainnet'
+rev = 'main'
```
davidiw, line 18: revert

core/sources/domain_e2e_tests.move
```diff
@@ -431,4 +502,124 @@ module aptos_names::domain_e2e_tests {
...
+            user,
+            option::none(),
+            test_helper::domain_name(),
+            test_helper::one_year_secs(),
+            test_helper::fq_domain_name(),
+            2,
+            vector::empty<u8>()
+        );
+    }
+    #[test(myself = @aptos_names, user = @0x077, aptos = @0x1, rando = @0x266f, foundation = @0xf01d)]
```
davidiw, line 607: nit new line

davidiw (approved):
let's get this landed!

core/sources/config.move
```diff
@@ -121,7 +123,7 @@ module aptos_names::config {
     /// The account will be used to manage names that are being used in a way that is harmful to others.
     /// Alternatively, the deployer can be used to perform admin actions.
     public fun signer_is_admin(sign: &signer): bool acquires ConfigurationV1 {
-        signer::address_of(sign) == admin_address() || signer::address_of(sign) == @aptos_names
+        signer::address_of(sign) == admin_address() || signer::address_of(sign) == @aptos_names || signer::address_of(sign) == @router_signer
```
movekevin, line 126: What is this router address? Why not reuse admin address?

core/sources/domains.move
```diff
@@ -403,18 +403,38 @@ module aptos_names::domains {
         register_name_internal(sign, subdomain_name, domain_name, registration_duration_secs, 0);
     }
 
+    public entry fun force_set_expiration(
```
movekevin, line 406: Is this going to be done for 1 domain at a time? Might be good to provide a batch version here as well

core/sources/domains.move
```diff
@@ -427,22 +447,43 @@ module aptos_names::domains {
...
-        time_is_expired(expiration_time_sec)
+        let now = timestamp::now_seconds();
+        if (expiration_time_sec > now) {
+            // Name has not expired. It is not available.
+            return false
+        };
+
+        // Name has expired.
+        if (option::is_some(&subdomain_name)) {
+            // We don't check registration grace period for subdomains. Name is available.
```
movekevin, line 467: This doesn't quite make sense to me product wise. Is this to allow people to register subdomains during the grace period?

movekevin: But then that would conflict with this line at the start of the function:

```
        if (option::is_some(&subdomain_name) && name_is_registerable(option::none(), domain_name)) {
```

because name_is_registerable(option::none(), domain_name)) would return false for the domain while in grace period. So name_is_registerable would return false for the subdomain as well

core/sources/domains.move
```diff
@@ -427,22 +447,43 @@ module aptos_names::domains {
...
+        };
+
+        // Name has expired AND it is a domain. Check registration grace period.
+        let expired_for = now - expiration_time_sec;
+        return expired_for > config::reregistration_grace_sec()
+    }
+
+    /// Returns true if the name is not registered OR (is registered AND is expired)
+    public fun name_is_expired(subdomain_name: Option<String>, domain_name: String): bool acquires NameRegistryV1 {
+        if (!name_is_registered(subdomain_name, domain_name)) {
```
movekevin, line 478: This code is weird. This seems to only differ from name_is_registered in that it doesn't take into account the grace period. Why not create a common function that has a variable grace period and passes it 0 for name_is_expired and non-zero for name_is_registerable?

core/sources/domains.move
```diff
@@ -600,7 +641,7 @@ module aptos_names::domains {
 
     /// This is a shared entry point for clearing the address of a domain or subdomain
     /// It enforces owner permissions
-    fun clear_name_address(
+    public fun clear_name_address(
```
movekevin, line 644: Why do we want to make this public? We already have clear_domain_address and clear_subdomain_address which honestly are much easier and clearer to use (register_name should have had 2 separate functions for domain and subdomain as well instead of muddling them together, which makes it nearly impossible to read)

core/sources/domain_e2e_tests.move
```diff
@@ -431,4 +502,125 @@ module aptos_names::domain_e2e_tests {
...
+
+        // Set the time past the domain's expiration time + grace period
+        let (_, expiration_time_sec, _) = domains::get_name_record_v1_props_for_name(option::none(), test_helper::domain_name());
+        timestamp::update_global_time_for_test_secs(expiration_time_sec + config::reregistration_grace_sec() + 1);
+
+        // Not owner anymore, name has expired
+        let (is_owner, _token_id) = domains::is_owner_of_name(user_addr, option::none(), test_helper::domain_name());
+        assert!(!is_owner, 1);
+
+        // Register the domain again. Works because it's past the grace period
```
movekevin, line 596: Add checks for name_is_registerable and name_is_expired both returning true?

movekevin (commented):
I noticed is_owner_of_name doesn't take into account the grace period as it uses name_is_expired: https://github.com/aptos-labs/aptos-names-contracts/blob/6bed75596a061c70a4e5106f4c0a73d423cee256/core/sources/domains.move#L521. Is this intended? From a product perspective, wouldn't it make more sense that the prev owner is still the owner during the grace period as no one else can register?

movekevin:
Also how do prev owners renew the domain during grace period? register_domain checks name_is_registerable which returns false during the grace period

core/sources/domain_e2e_tests.move
```diff
@@ -431,4 +502,125 @@ module aptos_names::domain_e2e_tests {
...
+            test_helper::domain_name(),
+            test_helper::one_year_secs(),
+            test_helper::fq_domain_name(),
+            2,
+            vector::empty<u8>()
+        );
+    }
+
+    #[test(myself = @aptos_names, user = @0x077, aptos = @0x1, rando = @0x266f, foundation = @0xf01d)]
+    fun test_register_after_reregistration_grace(myself: &signer, user: signer, aptos: signer, rando: signer, foundation: signer) {
```
movekevin, line 567: There's no test for renewing during the grace period.

PR #122, opened 2023-09-06
Add functionality to clear v1 registration

BriungRi:


router/sources/target_address_tests.move
```diff
@@ -15,11 +15,15 @@ module router::target_address_tests {
         domain_name: String,
         subdomain_name: Option<String>
     ): Option<address> {
-        let (_property_version, _expiration_time_sec, target_addr) = aptos_names::domains::get_name_record_v1_props_for_name(
-            subdomain_name,
-            domain_name,
-        );
-        target_addr
+        if (!aptos_names::domains::name_is_registered(subdomain_name, domain_name)) {
```
0xaptosj, line 18: i think you also need to apply this change to https://github.com/aptos-labs/aptos-names-contracts/blob/bl/rew/router/sources/router.move#L688

0xaptosj (approved):

PR #123, opened 2023-09-07
Clear updates

BriungRi:
- Allow `router_signer` to do privileged actions in v1
- Delete clear_registration
- Add test for repeated registration/clear
- Have router use `force_clear_reigstration` instead of `clear_registration
  - Clear registration upon migration and registration

### DIFF
--- a/router/sources/router.move
+++ b/router/sources/router.move
@@ -681,11 +681,15 @@ module router::router {
         domain_name: String,
         subdomain_name: Option<String>
     ): Option<address> {
-        let (_property_version, _expiration_time_sec, target_addr) = domains::get_name_record_v1_props_for_name(
-            subdomain_name,
-            domain_name,
-        );
-        target_addr
+        if (!aptos_names::domains::name_is_registered(subdomain_name, domain_name)) {
+            option::none()
+        } else {
+            let (_property_version, _expiration_time_sec, target_addr) = aptos_names::domains::get_name_record_v1_props_for_name(
+                subdomain_name,
+                domain_name,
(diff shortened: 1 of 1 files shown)


angieyth:
if we need to use `force_` functions, should we call those from router, or should we move those functions to router?

router/sources/router.move
```diff
@@ -190,6 +190,8 @@ module router::router {
                 domain_name,
                 registration_duration_secs,
             );
+            // Clear the name in v1
+            domains::force_clear_registration(&get_router_signer(), option::none(), domain_name)
```
rtso, line 194: Are we also handling clearing the v1 primary name somewhere? I'd suggest handling it inside of `force_clear_registration`.

router/sources/router.move
```diff
@@ -361,8 +364,10 @@ module router::router {
             );
             let token_id = aptos_names::token_helper::latest_token_id(&tokendata_id);
 
-            // Clear the registration in v1. This will also clear the primary name if applicable
-            domains::clear_registration(user, subdomain_name, domain_name);
+            // Clear the primary name in v1
+            if (is_primary_name) {
+                domains::clear_reverse_lookup(user);
+            };
```
rtso, line 370: This could be cleaner if we handled it within `force_clear_registration`

rtso:
LGTM, I'll defer to @angieyth and @0xaptosj for the final stamp.

angieyth (approved):
looking good

0xaptosj (approved):

PR #127, opened 2023-09-08
make migration error clear

0xaptosj:
add 2 new error type for name already migrated and cannot migrate subdomain before migrate domain

### DIFF
--- a/router/sources/router.move
+++ b/router/sources/router.move
@@ -39,7 +39,7 @@ module router::router {
     const ENOT_NAME_OWNER: u64 = 8;
     /// Subdomain has not been migrated
     const ESUBDOMAIN_NOT_MIGRATED: u64 = 9;
-    /// Trying to migrate subdomain before migrate domain
+    /// Cannot migrate subdomain before migrate domain
     const ECANNOT_MIGRATE_SUBDOMAIN_BEFORE_MIGRATE_DOMAIN: u64 = 10;
 
     // == OTHER CONSTANTS ==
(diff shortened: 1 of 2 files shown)


router/sources/router.move
```diff
@@ -363,6 +365,14 @@ module router::router {
             );
             let token_id = aptos_names::token_helper::latest_token_id(&tokendata_id);
 
+            // Domain must migrate before subdomain, throw error if this is a subdomain but domain has not been migrated
+            if (option::is_some(&subdomain_name)) {
+                assert!(
+                    exists_in_v2(domain_name, option::none()) && is_name_owner(user_addr, domain_name, option::none()),
```
rtso, line 371: I think this error would apply even if the user is not owner of the domain. Do we need the `is_name_owner` check?

0xaptosj: i think user must be the domain owner to be able to migrate the subdomain, @angieyth @BriungRi i think so, right?

angieyth: I  think Renee means we should throw the error when the domain is not v2 (whether the user is the owner or not)

angieyth: but also I think you don't have to be the domain owner to migrate the subdomain(?) you need to be the subdomain owner to do so.

rtso: +1 to @angieyth, for example I created a domain and subdomain, then transferred the subdomain to someone else. They should have the power to migrate the subdomain.

angieyth (approved):

router/sources/router.move
0xaptosj: make sense, updated

PR #128, opened 2023-09-11
Add bulk_migrate_domain and bulk_migrate subdomain

BriungRi:
I think there are some  issues with how vector of optionals works in the SDK. Temporary workaround.

Tested with @kaw2k @ https://explorer.aptoslabs.com/account/0x53febacc40e549ced4132bf3c3313076c3a81c631c8deda28cad871e34f6de0b/modules/code/migrate?network=testnet

### DIFF
--- a/bulk_migrate/sources/migrate.move
+++ b/bulk_migrate/sources/migrate.move
@@ -1,5 +1,6 @@
 module bulk_migrate::migrate {
     use std::error;
+    use std::option;
     use std::option::Option;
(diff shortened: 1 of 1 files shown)


0xaptosj (approved):

0xaptosj:
i did managed to make it work with [vector of option string](https://github.com/0xaptosj/ans-ts/blob/master/src/bulk_migrate.ts#L35), but separating does make it cleaner

PR #130, opened 2023-09-13
Reset subdomains on domain re-registration

BriungRi:
- Add `registration_time_sec` that gets assigned to current time during registration
 - Add `is_subdomain_registered_before_domain`. This gets used when checking if a name [subdomain] has expired. `is_name_registerable` relies on `is_name_expired` so this also gets accounted for
 - Add test to ensure subdomains can be re-registered if the domain gets re-registered
 - Fix `v2_test_helper::register_name` to use `registration_duration_secs` as opposed to `expiration_time_sec` for subdomains (requires changes to usages of `register_name` on subdomains`

### DIFF
--- a/core_v2/sources/v2_domains.move
+++ b/core_v2/sources/v2_domains.move
@@ -90,6 +90,7 @@ module aptos_names_v2::v2_domains {
         expiration_time_sec: u64,
         target_address: Option<address>,
         transfer_ref: object::TransferRef,
+        registration_time_sec: u64,
         // Currently unused, but may be used in the future to extend with more metadata
         extend_ref: object::ExtendRef,
         // Only present for subdomain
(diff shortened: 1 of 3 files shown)


core_v2/sources/v2_domains.move
```diff
@@ -1013,32 +1015,42 @@ module aptos_names_v2::v2_domains {
...
         } else {
             // Name is expired but haven't passed grace period, so name is not registerable
             return false
         }
     }
 
-    /// Returns true if the is not registered OR (name is registered AND is expired)
+    /// Returns true if
+    /// 1. The name is not registered OR
+    /// 2. The name is a subdomain AND subdomain was registered before the domain OR
```
0xaptosj, line 1041: can we move 2 to its own function?

0xaptosj (approved):

PR #132, opened 2023-09-13
refactor is owner of name in v2 and remove query_helper

0xaptosj:
1. Remove `is_owner_of_name` in v2 because it's dual check, instead use `is_token_owner` and `is_name_expired`
2. Remove `query_helper`, we should let all read request go through router


0xaptosj:
maybe we should just delete query helper, if there's any one used, just add to the router, so everything go through router

core_v2/sources/v2_test_helper.move
```diff
@@ -112,18 +112,12 @@ module aptos_names_v2::v2_test_helper {
...
         };
 
-        let is_owner = v2_domains::is_owner_of_name(user_addr, subdomain_name, domain_name);
+        let is_owner = v2_domains::is_token_owner(user_addr, domain_name, subdomain_name);
+        let is_expired = v2_domains::is_name_expired(domain_name, subdomain_name);
         // TODO: Re-enable / Re-write
         // let (tdi_creator, tdi_collection, tdi_name, tdi_property_version) = token::get_token_id_fields(&token_id);
 
-        assert!(is_owner, 3);
-        if (is_subdomain) {
```
0xaptosj, line 120: why are we checking these anyway

core_v2/sources/v2_test_helper.move
```diff
@@ -112,18 +112,12 @@ module aptos_names_v2::v2_test_helper {
...
-            let subdomain_name_copy = subdomain_name;
-            let subdomain_name_extracted = option::borrow(&subdomain_name_copy);
-            assert!(query_helper::is_owner_of_subdomain_name(user_addr, *subdomain_name_extracted, domain_name), 103);
-        } else {
-            assert!(query_helper::is_owner_of_domain_name(user_addr, domain_name), 103);
-        };
```
0xaptosj, line 126: these are already covered by L120

angieyth:
I think we talked about removing query_helper at some point, but didn't really do so. +1 to remove it

PR #139, opened 2023-09-15
Move SubdomainExt to be its own resource in the object

BriungRi:


0xaptosj (approved):
is it because it's bad idea to use optional struct in general?

PR #144, opened 2023-09-15
[ans] split get_name_props

angieyth:
split the function into `get_target_address_from_name` and `get_expiration_sec_from_name`


0xaptosj (approved):
this is good stuff

PR #149, opened 2023-09-15
Use object for ANS v2 contract management

BriungRi:
1. Use `object` instead of `resource_account`
2. Move all the events into that one object
3. Put structs storing event handler into a resource group
4. Remove unnecessary `store` for the structs storing event handler

Addresses https://github.com/aptos-labs/aptos-names-contracts/pull/23/files#r1321628442

### DIFF
--- a/core_v2/sources/v2_domains.move
+++ b/core_v2/sources/v2_domains.move
@@ -4,7 +4,7 @@ module aptos_names_v2::v2_domains {
     use aptos_framework::aptos_coin::AptosCoin;
     use aptos_framework::coin;
     use aptos_framework::event;
-    use aptos_framework::object::{Self, Object};
+    use aptos_framework::object::{Self, Object, ExtendRef, TransferRef};
     use aptos_framework::timestamp;
     use aptos_names_v2::v2_config;
     use aptos_names_v2::v2_price_model;
@@ -83,9 +83,9 @@ module aptos_names_v2::v2_domains {
     #[resource_group(scope = global)]
     struct ObjectGroup { }
 
-    /// Tokens require a signer to create, so this is the signer for the collection
-    struct App has key {
-        extend_ref: object::ExtendRef,
+    /// Tokens require a signer to create and we want to store global resources. We use object to achieve both
+    struct DomainObject has key {
+        extend_ref: ExtendRef,
     }
 
     #[resource_group_member(group = aptos_names_v2::v2_domains::ObjectGroup)]
@@ -93,10 +93,10 @@ module aptos_names_v2::v2_domains {
         domain_name: String,
         expiration_time_sec: u64,
         target_address: Option<address>,
-        transfer_ref: object::TransferRef,
+        transfer_ref: TransferRef,
         registration_time_sec: u64,
         // Currently unused, but may be used in the future to extend with more metadata
-        extend_ref: object::ExtendRef,
+        extend_ref: ExtendRef,
     }
 
     #[resource_group_member(group = aptos_names_v2::v2_domains::ObjectGroup)]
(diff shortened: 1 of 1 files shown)


0xaptosj (approved):

core_v2/sources/v2_domains.move
```diff
@@ -108,23 +108,27 @@ module aptos_names_v2::v2_domains {
...
     }
 
+    #[resource_group_member(group = aptos_framework::object::ObjectGroup)]
     /// Holder for `RegisterNameEvent` events
-    struct RegisterNameEvents has key, store {
+    struct RegisterNameEvents has key {
         register_name_events: event::EventHandle<RegisterNameEvent>,
     }
 
+    #[resource_group_member(group = aptos_framework::object::ObjectGroup)]
```
davidiw, line 129: probably should make a dedicated resource group for this... since it is all colocated within the same address....

core_v2/sources/v2_domains.move
```diff
@@ -81,12 +81,15 @@ module aptos_names_v2::v2_domains {
...
     const ECANNOT_RENEW_NAME_THAT_IS_EXPIRED_AND_PAST_GRACE_PERIOD: u64 = 31;
 
+    #[resource_group(scope = global)]
+    struct ObjectGroup { }
+
     /// Tokens require a signer to create, so this is the signer for the collection
-    struct CollectionCapability has key, drop {
-        capability: SignerCapability,
+    struct App has key {
+        extend_ref: object::ExtendRef,
```
movekevin, line 89: nit: Import ExtendRef directly. Also name the struct something more meaningful than App

core_v2/sources/v2_domains.move
```diff
@@ -108,23 +111,27 @@ module aptos_names_v2::v2_domains {
...
-    struct SetReverseLookupEvents has key, store {
+    struct SetReverseLookupEvents has key {
         set_reverse_lookup_events: event::EventHandle<SetReverseLookupEvent>,
     }
 
+    #[resource_group_member(group = aptos_names_v2::v2_domains::ObjectGroup)]
     /// Holder for `SetTargetAddressEvent` events
-    struct SetTargetAddressEvents has key, store {
+    struct SetTargetAddressEvents has key {
         set_name_events: event::EventHandle<SetTargetAddressEvent>,
```
movekevin, line 123: Why separate events into so many structs?

core_v2/sources/v2_domains.move
```diff
@@ -191,44 +198,43 @@ module aptos_names_v2::v2_domains {
...
+        let constructor_ref = object::create_named_object(
             account,
-            APP_SIGNER_CAPABILITY_SEED,
+            APP_OBJECT_SEED,
         );
-        move_to(account, CollectionCapability {
-            capability: token_signer_cap,
-        });
+        let extend_ref = object::generate_extend_ref(&constructor_ref);
+        let app_signer = object::generate_signer(&constructor_ref);
```
movekevin, line 207: nit: let app_signer = &... to save the extra & everywhere else

core_v2/sources/v2_domains.move
```diff
@@ -191,44 +198,43 @@ module aptos_names_v2::v2_domains {
...
         collection::create_unlimited_collection(
-            &token_resource,
+            &app_signer,
             utf8(SUBDOMAIN_COLLECTION_DESCRIPTION),
             v2_config::subdomain_collection_name(),
             option::none(),
             utf8(COLLECTION_URI),
         );
+        aptos_account::create_account(signer::address_of(&app_signer));
+        move_to(&app_signer, SetTargetAddressEvents {
```
movekevin, line 223: Why not combine all of these into the same resource?

core_v2/sources/v2_domains.move
movekevin: If most of these functions that emit different events change NameRecord, the conflict would still happen anyway right?

PR #151, opened 2023-09-15
Remove unused config in v2

0xaptosj:


0xaptosj:
this covers https://github.com/aptos-labs/aptos-names-contracts/pull/150

PR #171, opened 2023-09-22
Update unit test to check primary name is always cleared in v1 after set in v2

0xaptosj:
Whenever primary name is set in v2, it should be cleared in v1 because of [this line](https://github.com/aptos-labs/aptos-names-contracts/blob/bl/rew/router/sources/router.move#L522), adding unit test to make sure of that

### DIFF
--- a/router/sources/tests/migration_tests.move
+++ b/router/sources/tests/migration_tests.move
@@ -154,11 +154,17 @@ module router::migration_tests {
         // Make v1 read only except for admin
         aptos_names::config::set_is_enabled(aptos_names, false);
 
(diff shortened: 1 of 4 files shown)


router/sources/tests/migration_tests.move
```diff
@@ -154,11 +154,17 @@ module router::migration_tests {
         // Make v1 read only except for admin
         aptos_names::config::set_is_enabled(aptos_names, false);
 
-        // Migration fails because user2 does not own `domain_name`
```
0xaptosj, line 157: this is copied by mistake

angieyth (approved):

PR #177, opened 2023-10-04
Add bulk register subdomain

BriungRi:


bulk/sources/bulk.move
```diff
@@ -78,4 +78,33 @@ module bulk::bulk {
...
+        domain_admin: &signer,
+        domain_names: vector<String>,
+        subdomain_names: vector<String>,
+        expiration_time_secs: vector<u64>,
+        expiration_policies: vector<u8>,
+        transferrable: vector<bool>,
+        target_addrs: vector<address>,
+        to_addrs: vector<address>,
+    ) {
+        let idx = 0;
```
angieyth, line 95: do you think we should provide so many options at once? like different `to_addr` for each subdomain? if so , do you think we should add a check to make sure every vector has same length?

angieyth (approved):
do we have any request for this function?

0xaptosj (approved):

PR #184, opened 2024-05-09
Fix router in post cutoff period

BriungRi:
Two bugs:

1. Originally, if the name is expired but within the grace period, we don't consider the token to be owned anymore. This meant we did not auto-migrate. The first fix is to consider the token owned while in the grace period.
2. Before the cutoff, we would always give users an extra year before migration, which meant we could guarantee users got a positive `registration_duration_secs`. However, since we are now past the cutoff, we can find ourselves in a situation where the new expiration (e.g. 15 days ago) is smaller than `now_seconds`, giving us a negative `registration_duration_secs`. We have decided (Alyssa was looped in) to simply zero out the days in the grace period (15 days into grace period, give them a free 15 days upon migration so their new expiration is `now` + 1 second) which unblocked renewals.

This issue is temporary because v1 names will all expire soon, meaning the case where a user needs to renew a v1 name will disappear soon.

### DIFF
--- a/sh_scripts/dev_setup.sh
+++ b/sh_scripts/dev_setup.sh
@@ -11,10 +11,8 @@ if ! command -v aptos &>/dev/null; then
     echo "aptos could not be found"
     echo "installing it..."
     TARGET=Ubuntu-x86_64
-    VERSION=2.0.2
+    VERSION=3.3.0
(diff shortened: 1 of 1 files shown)


davidiw:
Fix tests and land please

PR #191, opened 2024-11-13
Fix subdomain name expiration for primary names

BriungRi:
# Summary

Primary names will expire if the name itself is expired. Previously, the expiration logic only looked at the record's expiration timestamp, ignoring cases where there was a subdomain policy to follow the domain's expiration. This PR fixes that

# Test Plan

Add a test that checks the exact scenario described above


core_v2/sources/v2_1_domains.move
```diff
@@ -715,7 +715,7 @@ module aptos_names_v2_1::v2_1_domains {
     /// Returns the reverse lookup (the token addr) for an address if any.
     public fun get_reverse_lookup(
```
davidiw, line 716: could probably be a view function...

